Issue #3, opened 2026-06-15
fix: wire config_freeze.rs into set_config() authorization path


usmanimamu17-create:
## Overview

`config_freeze.rs` implements freeze/unfreeze but these functions are never called from `set_config()`. The freeze flag has zero gating effect.

## Evidence

**File:** `apexchainx_calculator/src/config_freeze.rs` — implements freeze logic
**File:** `apexchainx_calculator/src/lib.rs` — set_config() has no freeze check

## Impact
- Admin who calls freeze_config() has false sense of security
- Config can still be changed despite freeze flag being set

## Proposed Solution
Add require_not_frozen() check in set_config(). Expose freeze/unfreeze as admin-gated methods. Emit events.

## Acceptance Criteria
- [ ] set_config() returns error when config is frozen
- [ ] Error uses clear variant (e.g., ConfigFrozen = 16)
- [ ] Admin can freeze/unfreeze via contract methods
- [ ] Freeze/unfreeze emit versioned events
- [ ] is_config_frozen() returns correct state
- [ ] Unfrozen state allows normal updates
- [ ] All existing tests pass

## File Map
- `apexchainx_calculator/src/config_freeze.rs`
- `apexchainx_calculator/src/lib.rs`
- `apexchainx_calculator/src/tests.rs`

## Testing Strategy
Unit: Freeze → attempt set_config → verify rejection. Unfreeze → verify success.

## Estimated Effort: 3h

aliyuHabibu:
Hi, I'm a software engineer, I've gone through the issue and I can resolve it. Assign it to me thank you

PR #26, opened 2026-06-20
fix: update threshold_config set_config calls

Errordog2:
## Summary
- updates `threshold_config.rs` to use the current `set_config(caller, severity, threshold_minutes, penalty_per_minute, reward_base)` API
- removes the now-unused `SLAConfig` import
- keeps the existing threshold test assertions and scenarios unchanged

## Validation
- `git diff --check` passes locally
- `cargo test -p apexchainx_calculator threshold_config --no-default-features` could not complete in this Windows environment because the MSVC linker `link.exe` is not installed


usmanimamu17-create:
Thanks for the cleanup @Errordog2! Quick heads-up before I can merge this: per our contribution guidelines, every PR needs to be linked to an approved tracking issue so maintainers have the context to triage. I don't see a `Closes #…` (or `Fixes #…`) reference on this PR, and the body doesn't mention which problem this solves. Could you open or link an issue describing the motivation (e.g. test compilation breakage on the new set_config signature) so reviewers and future contributors can follow the *why*? Holding off on merging until that's in place.

usmanimamu17-create:
Hi @Errordog2, thanks for the PR. The closest matching issue (#3) is assigned to @aliyuHabibu, so under our author=assignee policy we can't merge this from your handle. Also looks like #56 by @Joeloo1 already addresses the same threshold_config set_config concern. Could you coordinate on the issue, or close this in favor of the existing PR? 🙏

PR #47, opened 2026-06-20
test: cover normalized history flags

### LINKED ISSUE #38: test: cover normalize_history has_violations and has_rewards flags
`history_snapshot.rs` exposes `NormalizedSnapshot` with `has_violations` and `has_rewards` aggregate flags, but the existing test in the file asserts only that `client.get_stats().total_calculations == 2`. The boolean outputs (the whole point of `normalize_history`) are never exercised. Add tests that submit both a violation (e.g. `mttr > threshold`) and a reward (e.g. `mttr << threshold`) for the same severity, call `normalize_history`, and assert `has_violations == true`, `has_rewards == true`, `count == 2`. Also add a control case where two met-rated entries yield `has_rewards == true` and `has_violations == false`.

9904099:
## Description
Adds focused coverage for `normalize_history` aggregate flags so backend consumers can rely on the snapshot booleans, not just the raw calculation count.

## Type of Change
- [x] Documentation update
- [x] Test update

## Related Issues
Fixes #38

## Changes Made
- Expose the existing `history_snapshot` module from the crate so its tests compile and run with the contract test suite.
- Add a mixed history case with one reward and one violation, asserting `count == 2`, `has_rewards == true`, and `has_violations == true`.
- Add a control case with two met entries, asserting `count == 2`, `has_rewards == true`, and `has_violations == false`.

## Testing
- [x] Unit tests added/updated

```sh
cargo fmt --manifest-path apexchainx_calculator/Cargo.toml -- --check
cargo test --manifest-path apexchainx_calculator/Cargo.toml history_snapshot -- --nocapture
```

The targeted test run passes with the repository's existing warnings in unrelated modules.

Contact: 50889616@qq.com

### DIFF
--- a/apexchainx_calculator/src/history_snapshot.rs
+++ b/apexchainx_calculator/src/history_snapshot.rs
@@ -16,8 +16,8 @@
 //! The normalization is deterministic: identical history inputs always produce
 //! identical snapshot outputs.
 
-use soroban_sdk::{symbol_short, Vec};
 use crate::SLAResult;
+use soroban_sdk::{symbol_short, Vec};
 
 /// Summarised view of SLA calculation history.
 ///
@@ -59,8 +59,9 @@ pub fn normalize_history(history: &Vec<SLAResult>) -> NormalizedSnapshot {
 
 #[cfg(test)]
 mod tests {
-    use soroban_sdk::{symbol_short, testutils::Address as _, Address, Env};
+    use super::normalize_history;
     use crate::{SLACalculatorContract, SLACalculatorContractClient};
+    use soroban_sdk::{symbol_short, testutils::Address as _, Address, Env};
 
     #[test]
     fn test_history_snapshot_is_deterministic() {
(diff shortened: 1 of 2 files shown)


9904099:
Thanks for checking this. I re-ran the validation locally and the change itself is still passing on my side.

Local validation:

```text
cargo fmt --manifest-path apexchainx_calculator/Cargo.toml -- --check
cargo clippy --manifest-path apexchainx_calculator/Cargo.toml
cargo test --manifest-path apexchainx_calculator/Cargo.toml history_snapshot -- --nocapture
cargo test --manifest-path Cargo.toml --all -- --nocapture
```

The GitHub Actions failure appears to be infrastructure/network related rather than a test failure. The job fails while `cargo clippy` is downloading the crates.io dependency `syn`, with the connection reset during download. The new `history_snapshot` tests do not run before that failure point.

I am leaving the PR unchanged for now so the diff stays limited to the issue scope. If you would prefer, I can re-run the workflow after the registry/network issue clears.

usmanimamu17-create:
Thanks for the contribution @9904099! Unfortunately the `Build, Test & Verify` check is currently failing on this PR (run https://github.com/ApexChainx/ApexChainx-Contracts/actions/runs/27872255678 ). Could you take a look at the workflow logs, fix the failure, and push a follow-up commit so we can re-run CI and proceed with the review? Happy to merge once green.

9904099:
Thanks for the heads-up. I checked the workflow log again, and the failing step still stops before clippy can evaluate this PR's Rust code:

```text
cargo clippy --manifest-path apexchainx_calculator/Cargo.toml
error: failed to get `syn` as a dependency of package `bytes-lit v0.0.5`
...
download of 3/s/syn failed
curl failed
[56] Failure when receiving data from the peer (Recv failure: Connection reset by peer)
```

So this looks like the same crates.io registry/network failure rather than a code or test failure in the PR.

I pushed an empty commit (`cc9aad2`) to retrigger CI without changing the implementation. If the rerun surfaces an actual compile, clippy, or test failure after dependencies download successfully, I will fix that directly.

usmanimamu17-create:
Hi @9904099, thanks for putting this together! Per our policy the PR author should match the assignee on the linked issue — and #38 is currently assigned to @oluwaseyi1996-netizen. Could you coordinate with them, or request reassignment on the issue, before we can merge? 🙏

PR #48, opened 2026-06-20
fix: implement rollback and finalize methods for cross-contract safety

MerlinTheWhiz:
## Description

Implements the missing `rollback_all()` and `finalize()` methods on `CrossContractSafety`, and stores the target contract `Address` in the compensation stack so that compensations can actually be invoked during rollback. Previously the module-level docs and `call()` API comments referenced these methods without providing them, leaving the compensation stack populated on success and never drained — the documented rollback guarantee was unfulfilled.

## Type of Change
- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Related Issues
Fixes #32

## Changes Made
- Added `Address` to imports in `cross_contract_safety.rs`
- Changed `compensation_stack` type from `Vec<(Symbol, CompensationAction)>` to `Vec<(Address, Symbol, CompensationAction)>` to store the target contract ID for each registered compensation
- Updated `call()` to store `contract_id` in the compensation stack tuple
- Implemented `rollback_all()` — walks the compensation stack in reverse order, invoking each compensation action on its target contract via `env.try_invoke_contract`, then drains the stack
- Implemented `finalize()` — clears the compensation stack on the happy path (no rollback needed)
- Updated module-level documentation and `call()` doc comment to accurately reflect the API
- Updated `coordination_harness.rs` test to use the new 3-tuple format

## Testing
- [x] Unit tests added/updated
- [ ] Integration tests added/updated
- [x] Manual testing performed

All 392 existing tests continue to pass. Added 4 new tests:
- `test_rollback_all_clears_stack` — verifies that `rollback_all()` drains the stack
- `test_rollback_all_empty_stack_is_noop` — verifies no-op on empty stack
- `test_finalize_clears_stack` — verifies that `finalize()` drains the stack
- `test_finalize_empty_stack_is_noop` — verifies no-op on empty stack

## Checklist
- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my own code
- [ ] I have commented complex logic
- [x] I have updated relevant documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] My changes do not introduce new warnings

## Screenshots
image

Issue #32, opened 2026-06-20
fix: cross_contract_safety references non-existent rollback_all and finalize methods


usmanimamu17-create:
The module-level documentation and the `call()` API comments in `apexchainx_calculator/src/cross_contract_safety.rs` refer to a `safety.rollback_all()` and a `safety.finalize()` step, but neither method exists on `CrossContractSafety`. The compensation stack is populated on success and never drained, so any partial-failure scenario leaks registered compensations and the documented rollback guarantee is unfulfilled. Implement `rollback_all` (walks the stack in reverse invoking the registered compensation actions) and `finalize` (confirms the happy path commits). Until they exist, the file promises a safety guarantee it cannot deliver.

MerlinTheWhiz:
Hello, I'm MerlinThewhiz, a software engineer with a strong background in smart contract, and a couple years in my arsenal. Having read the task requirement, I'd love to work on this issue. Please assign.

MerlinTheWhiz:
Hi @Icahbod thanks for the opportunity to contribute. I've attached a PR, ready for review.

PR #50, opened 2026-06-20
Fix/sla config version idempotency

### LINKED ISSUE #29: fix: calculate_sla duplicate detection ignores config_version_hash
When `calculate_sla` detects an existing entry for an `outage_id`, it only compares `mttr_minutes` and `threshold_minutes` to decide idempotency. The `config_version_hash` stamped on the prior result is never compared, so a duplicate request submitted after a `set_config` change silently returns the old result and rejects updated inputs as `DuplicateOutageInput`. The doc comment promises the duplicate check ensures "execution inputs resolve to the same deterministic result"; extend the comparison to include `config_version_hash` so the contract honours its documented invariant and treats a config-bumped duplicate as a fresh calculation rather than as a conflict.

MerlinTheWhiz:
## Description

Closes #29 

Fix `calculate_sla` duplicate detection to compare `config_version_hash` in addition to `mttr_minutes` and `threshold_minutes`. Previously, submitting a duplicate `outage_id` after a `set_config` change would silently return the stale cached result, ignoring that the config (and therefore the deterministic result) had changed.

The doc comment on the duplicate check promises idempotency only when "execution inputs resolve to the same deterministic result." Since `config_version_hash` is part of those inputs, the comparison must include it to honour the contract invariant.

## Type of Change
- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation update

## Related Issues
Relates to #29 

## Changes Made
- **`apexchainx_calculator/src/lib.rs`**: In `calculate_sla`, wrap the existing `mttr_minutes`/`threshold_minutes` duplicate check inside a `config_version_hash` equality guard. If the hashes differ (config was updated), execution falls through to the fresh calculation path — the new result is appended to history, stats are incremented, and events are emitted. If hashes match, the existing logic applies: same inputs → idempotent return, different inputs → `DuplicateOutageInput` error.
- **`apexchainx_calculator/src/tests.rs`**: Added 4 tests:
  - `test_config_bumped_duplicate_treated_as_fresh_calculation` — reward_base changed, same outage_id/mttr/threshold → recalculates with new config, history has 2 entries, stats show 2 calculations
  - `test_config_bumped_duplicate_threshold_change_is_fresh` — threshold lowered, same outage_id/mttr → result flips from met→viol
  - `test_duplicate_same_config_still_idempotent` — no config change → identical duplicate still returns cached result
  - `test_duplicate_same_config_with_different_mttr_still_panics` — same config, different mttr → still errors with `DuplicateOutageInput`

## Testing
- [x] Unit tests added/updated
- [x] Manual testing performed

All 393 tests pass (3 slow stress tests skipped). `cargo fmt --all --check` and `cargo clippy --all-targets` produce no new warnings.

## Checklist

## Screenshot (the 4 new added tests)
image

### DIFF
--- a/apexchainx_calculator/src/coordination_harness.rs
+++ b/apexchainx_calculator/src/coordination_harness.rs
@@ -227,7 +227,9 @@ mod coordination_harness_tests {
 
         // Simulate two successful calls by directly registering compensations
         // (the actual call path would register them via safety.call())
+        let mock_contract = Address::generate(&env);
         safety.compensation_stack.push_back((
+            mock_contract.clone(),
             symbol_short!("lock_fnds"),
             cross_contract_safety::CompensationAction {
                 tag: cross_contract_safety::COMP_UNLOCK_FUNDS,
@@ -235,6 +237,7 @@ mod coordination_harness_tests {
             },
         ));
         safety.compensation_stack.push_back((
+            mock_contract,
             symbol_short!("rel_pay"),
             cross_contract_safety::CompensationAction {
                 tag: cross_contract_safety::COMP_REVERSE_SETTLE,
--- a/apexchainx_calculator/src/cross_contract_safety.rs
+++ b/apexchainx_calculator/src/cross_contract_safety.rs
@@ -192,16 +192,12 @@ impl CrossContractSafety {
     /// This should be called after a fatal error to undo all prior
     /// successful cross-contract calls.
     pub fn rollback_all(&mut self, env: &Env) {
-        while let Some((contract_id, _function_name, action)) = self.compensation_stack.pop_back()
-        {
+        while let Some((contract_id, _function_name, action)) = self.compensation_stack.pop_back() {
             // Invoke the compensation function on the original target contract.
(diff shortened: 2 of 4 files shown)


usmanimamu17-create:
Excellent fix @MerlinTheWhiz! Your analysis of the idempotency invariant is spot-on — the doc-comment contract clearly says idempotency holds when *execution inputs resolve to the same deterministic result*, and `config_version_hash` is part of those inputs, so wrapping the existing `mttr_minutes`/`threshold_minutes` check in a config-hash guard was the right move. The four new tests pin down exactly the right matrix: bumped-config → fresh calculation with history+stats updated; threshold flip → met/viol reversal; same config + same inputs → cached idempotent; same config + different inputs → `DuplicateOutageInput`. 393/393 passing with no new clippy warnings — clean. Note: PR #48 (the rollback/finalize work for #32) has already landed on main, and the cross_contract_safety.rs / coordination_harness.rs changes here are byte-identical to those, so they apply as a no-op. Merging this one too.

PR #64, opened 2026-06-23
fix migration harness Issue

### LINKED ISSUE #34: fix: migration harness cannot initialise new storage fields on bump
`migrate()` in `apexchainx_calculator/src/lib.rs` includes only a `v0 → v1` arm that stamps the version key. The doc comment promises future arms as placeholder comments, but the harness does not codify the contract for adding a migration step. When `STORAGE_VERSION` is bumped (say to 2) and a new field is added to an existing struct (e.g. a per-severity retention counter on `SLAConfig`), a struct read that gains a non-optional field will fail to deserialise outright; a fall-back `unwrap_or(default)` will silently zero the new value. Codify the migration contract in the harness: every new `STORAGE_VERSION` bump must (a) initialise every newly-added storage key with a deterministic default and (b) add a regression test that exercises the migration on a frozen `v(N-1)` snapshot, before the next breaking release lands.

abk-tech:
Summary

Fixed an issue where the migration harness could not properly initialize newly added storage fields during version bumps.

Changes
Updated migration initialization logic to correctly handle new storage fields.
Ensured storage state remains consistent across migrations and upgrades.
Added validation for newly introduced fields during migration execution.
Outcome

New storage fields are now initialized correctly during migration bumps, preventing migration failures and ensuring smooth upgrades.

Closes: #34

### DIFF
--- a/apexchainx_calculator/src/lib.rs
+++ b/apexchainx_calculator/src/lib.rs
@@ -546,6 +546,71 @@ impl SLACalculatorContract {
         Ok(())
     }
 
+    // Initialise any storage keys that may be missing from older schema
+    // versions. This is intentionally conservative: only set a value when
+    // the key is absent so migration is idempotent and does not overwrite
+    // existing state.
+    fn init_missing_storage_defaults(env: &Env) {
+        let inst = env.storage().instance();
+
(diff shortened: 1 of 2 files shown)


usmanimamu17-create (approved):
LGTM

PR #124, opened 2026-07-15
feat: add workflow secret-leak check (#122)

### LINKED ISSUE #122: Sanity-check CI never echoes secrets
**Problem Statement.** No static check verifies that workflow YAML
doesn't `echo ${{ secrets.X }}` or `run: env`.

**Expected Outcome.** Workflow lint step that searches workflow files
for unsafe patterns (e.g., `echo.*secrets`). Fail CI on hit.

**Files:** New `tools/secret-leak-check.ts`.

neromtoobad:
## Summary

Closes #122. Adds a static lint so workflow YAML can never print secret material into the build log, and fails CI on a hit.

GitHub Actions masks known secret values, but masking is best-effort — it breaks on transformed values (base64, substrings, JSON-encoded) and does nothing about a full environment dump. The cheapest defence is to never print the secret at all.

## Files

**`tools/secret-leak-check.ts`** (new) — scans `.github/workflows/*.yml|yaml`, exits non-zero on a hit. Follows the existing `tools/` conventions (`ts-node` shebang, CommonJS, usage header).

| Rule | Catches |
|---|---|
| `echo-secrets` | `echo`/`printf` of a `${{ secrets.* }}` expression |
| `env-dump` | a bare `env` / `printenv` shell command |
| `echo-secret-env` | `echo`/`printf` of a shell var the same workflow mapped from `${{ secrets.* }}` (the indirect leak) |

A reviewed false positive can be suppressed inline: `# secret-leak-check:allow <reason>`.

**`.github/workflows/workflow-lint.yml`** (new) — runs it on every push/PR to `main`.

## Two decisions worth flagging

1. **`env:` vs `env`.** The `env:` YAML mapping key is legitimate and everywhere, so the `env-dump` rule only matches `env`/`printenv` as a *shell command* and explicitly excludes anything ending in `:`. Without that, the rule would false-positive across the whole repo.

2. **Own workflow instead of a step in `ci.yml`.** I initially wired this into `ci.yml`, then found **`ci.yml` currently fails to parse as YAML** — line 119, `run: cargo test --lib fuzz_tests::` (both `js-yaml` and Ruby `psych` reject it; GitHub lists those runs as `.github/workflows/ci.yml` rather than `name: CI`, which is what it shows when it can't parse the file). A step added there would never run, so the lint lives in its own workflow — which also means no Rust toolchain and a seconds-long job. **This PR does not touch `ci.yml`** (that parse bug looks like a separate issue worth its own fix — happy to open one).

## Testing

- **Passes clean on this repo as-is:** 4 workflow files scanned, 0 findings, exit 0 — including its own new workflow.
- **Negative fixture** (direct secret echo, `env`, `printenv | grep`, indirect secret-derived echo) → exactly **4 findings, exit 1**.
- **No false positives:** an allowlisted line, the legitimate `env:` blocks, and a plain `echo "building"` are all correctly ignored.
- `workflow-lint.yml` validated as YAML 1.2.

Happy to adjust the rule set, add more patterns, or fold it into `ci.yml` instead once that file parses.

### DIFF
--- a/.github/workflows/workflow-lint.yml
+++ b/.github/workflows/workflow-lint.yml
@@ -1,0 +1,48 @@
+# =============================================================================
+# ApexChainx Contracts — Workflow Secret Lint (issue #122)
+# =============================================================================
+#
+# Static check that no workflow writes secret material into the build log.
+# GitHub Actions masks known secret values, but masking is best-effort: it
+# breaks on transformed values (base64, substrings, JSON-encoded) and does
+# nothing about a full environment dump. The cheapest defence is to never print
+# the secret at all — this lints the workflow YAML for the patterns that do and
+# fails CI on a hit.
+#
+# Kept as its own workflow (rather than a step in ci.yml) so it needs no Rust
+# toolchain, runs in seconds, and reports independently of the contract build.
+#
+# Run locally:  npx --yes tsx tools/secret-leak-check.ts
+# =============================================================================
+
+name: Workflow Secret Lint
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+concurrency:
+  group: workflow-secret-lint-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  secret-leak-check:
+    name: Secret Leak Check
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+
+      - name: Check workflows never echo secrets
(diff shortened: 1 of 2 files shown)


usmanimamu17-create:
Thanks for adding the secret-leak workflow! Just a heads-up though: the "Secret Leak Check" you can see on the Checks tab is actually the workflow this very PR introduces — its green run just confirms the new workflow file parses and executes, not that any contract CI has been exercised.

The actual merge gate for this repo is `ci.yml`: `client-checks` (build / WASM no-std / format / clippy), `e2e-tests`, `fuzz-tests` (proptest), and `provenance-hashes` (WASM release + SHA). None of them have registered a run on this PR. Could you rebase onto latest `main` and push once more so the four PR-gating jobs fire? Once they are green I will squash-merge. 🚀

Issue #93, opened 2026-07-15
Custom severity-level support via `set_custom_severity`


usmanimamu17-create:
**Problem Statement.** Only four hardcoded severities (`critical`,
`high`, `medium`, `low`). Production deployments may want
`warning`, `info`, `critical_security`, `service_degraded`, etc.

**Outcome-Specific Constraints (must hold for the implementation PR):**
1. `RESULT_SCHEMA_VERSION` MUST NOT bump (custom severities are an additive extension; no breakage).
2. Canonical severity ordering (used by `get_config_snapshot`, `compute_config_version_hash`, and `get_result_schema`) MUST preserve the existing `critical/high/medium/low` order. Custom severities MUST be appended **after** the canonical four in a deterministic alphabetical or insertion-ordered fashion — see #61.
3. The four canonical validators in `validate_config` are preserved unchanged. Custom severities use only the general bounds (`threshold <= 1440`, `penalty in 1..10000`, `reward in 1..100000`).
4. `get_failure_schema` extended with a new `SLAError::SeverityNotInSet` variant (see #23) when a custom severity is referenced but not registered.

**Expected Outcome.**
- New `Map<Symbol, SLAConfig>` storage slot, distinct from the four canonical entries (avoid `get_config_snapshot` returning canonical + custom as one mixed list).
- `set_custom_severity(caller, severity)` adds the key with default
  values; `remove_custom_severity(caller, severity)` removes it.
- 4 hardcoded defaults loaded at init.
- Update `get_config_snapshot` to iterate custom map.
- Test: add `"warning"` severity dynamically, run calculate, get result.

**Files:** `lib.rs`, all event-correlated tests.

Yasir-TechGuy:
Please assign this issue to me

Yasir-TechGuy:
@Icahbod Quick clarification before I start: the Expected Outcome says both "avoid get_config_snapshot
returning canonical + custom as one mixed list" and "Update get_config_snapshot to iterate
custom map." I'm reading this as: get_config_snapshot() itself stays untouched (canonical-only,
preserving RESULT_SCHEMA_VERSION and existing ordering), and I'll add a new parallel function
(e.g. get_custom_config_snapshot()) that iterates the custom map the same way. Let me know if
that's not what you intended.

PR #340, opened 2026-07-29
test(apexchainx_calculator): add failure catalog drift guard test

### LINKED ISSUE #264: Add a `failure catalog drift` test so the error helper layer and the public error enum cannot diverge
Problem Statement
The contract exposes an error enum and an external helper layer, but there is no dedicated regression test proving they remain synchronized over time.

Why it matters
An error helper drift is an easy way to create surprising backend behavior and inaccurate status handling.

Technical Context
The error enum is in [apexchainx_calculator/src/lib.rs](apexchainx_calculator/src/lib.rs), while the helper layer is in [apexchainx_calculator/src/error_responses.rs](apexchainx_calculator/src/error_responses.rs).

Expected Outcome
Introduce a drift guard test that ensures all public error utilities remain aligned with the contract enum.

Acceptance Criteria
- A drift guard test exists.
- The test fails if a helper is added or removed without the enum being updated.
- It runs in the normal contract test suite.

Implementation Notes
This is a small but high-value regression check that strengthens the contract surface.

Files or modules likely to be affected
- [apexchainx_calculator/src/lib.rs](apexchainx_calculator/src/lib.rs)
- [apexchainx_calculator/src/error_responses.rs](apexchainx_calculator/src/error_responses.rs)
- [apexchainx_calculator/src/tests.rs](apexchainx_calculator/src/tests.rs)

Dependencies (if any)
No direct dependency.

Difficulty level
Medium

Estimated effort
M

paul-motron:
## Summary
- Adds a compile-time drift guard test that binds every `SLAError` variant to its `error_responses::is_*` predicate through an exhaustive match with no wildcard arm — a new/removed enum variant or a renamed/removed helper fails the build, which fails the contract test suite.
- Adds a mutual-exclusivity test so a copy-pasted helper body (matching the wrong variant) is also caught, not just missing/extra arms.
- Fixes the drift that already existed: `SLAError::SeverityNotInSet` (added for #93) had no corresponding `error_responses::is_severity_not_in_set` helper. Added it, in enum order, alongside the other predicates.

## Why
The error enum lives in `apexchainx_calculator/src/lib.rs` and the backend-facing helper layer lives in `apexchainx_calculator/src/error_responses.rs`, but nothing enforced that they stay in sync. `SeverityNotInSet` had already drifted (present in the enum and in `get_failure_schema`'s catalogue, but missing its helper) — this PR both fixes that gap and adds a regression test so it can't silently happen again.

## Test plan
- [x] `cargo test test_failure_catalog` — both new tests pass
- [x] Manually removed `is_severity_not_in_set` and confirmed the crate fails to compile (`error[E0425]: cannot find function is_severity_not_in_set in module error_responses`), proving the guard actually catches drift
- [x] `cargo test` — full suite passes (493 passed; 0 failed)

Closes #264

### DIFF
--- a/apexchainx_calculator/src/error_responses.rs
+++ b/apexchainx_calculator/src/error_responses.rs
@@ -68,6 +68,10 @@ pub fn is_invalid_input(err: &SLAError) -> bool {
     matches!(err, SLAError::InvalidInput)
 }
 
+pub fn is_severity_not_in_set(err: &SLAError) -> bool {
+    matches!(err, SLAError::SeverityNotInSet)
+}
+
 pub fn is_outage_recalc_limit(err: &SLAError) -> bool {
     matches!(err, SLAError::OutageRecalcLimit)
 }
--- a/apexchainx_calculator/src/tests.rs
+++ b/apexchainx_calculator/src/tests.rs
@@ -7284,4 +7284,151 @@ fn test_historical_parity_golden_results() {
     let schema = client.get_result_schema();
     assert_eq!(schema.schema_version, 1, "Result schema version changed — check migration notes");
     assert_eq!(schema.deprecated_symbols.len(), 0, "Unexpected deprecated symbols in v1");
+// #264 – Failure catalog drift guard
+// ============================================================
+//
+// `error_responses.rs` provides a `is_*` predicate for every `SLAError`
+// variant so backend consumers never have to match on the enum directly.
+// The match below binds each variant to its predicate with no wildcard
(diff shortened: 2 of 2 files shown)


usmanimamu17-create (approved):
LGTM! 🚀

usmanimamu17-create (approved):
Nice drift guard! 🛡️

usmanimamu17-create (approved):
Nice drift guard! 🛡️

usmanimamu17-create:
Merged — clean work, thanks! 🙌

Issue #271, opened 2026-07-23
Document how the repository’s `SC-` feature markers should be used during future module development


usmanimamu17-create:
Problem Statement
The repository uses many `SC-` and `SC-W5-` markers in code comments and tests, but their maintenance lifecycle is not clearly documented.

Why it matters
Without a policy, these markers may become stale or incomplete, reducing their value as design history.

Technical Context
The markers appear throughout the contract crate, especially in [apexchainx_calculator/src/lib.rs](apexchainx_calculator/src/lib.rs) and the specialized test modules.

Expected Outcome
Publish a contribution note describing how maintainers should add, update, or retire `SC-` markers.

Acceptance Criteria
- A marker policy exists in the repo.
- The doc explains the intended purpose and expected review discipline.
- Examples are linked from the docs.

Implementation Notes
Keep the policy discipline-oriented and simple to follow.

Files or modules likely to be affected
- [docs/PROJECT_CONTEXT.md](docs/PROJECT_CONTEXT.md)
- [CONTRIBUTING.md](CONTRIBUTING.md)
- [apexchainx_calculator/src/lib.rs](apexchainx_calculator/src/lib.rs)

Dependencies (if any)
No direct dependency.

Difficulty level
Low

Estimated effort
S

Kureszn:
@Kureszn has applied to work on this issue as part of the **Stellar Wave Program**'s 7th wave.



> I've reviewed the issue and believe I can contribute effectively. My experience spans software development, blockchain technologies, cybersecurity and open source collaboration. I take a structured approach to contributions carefully understanding the existing codebase, identifying the root cause and delivering a thoroughly tested implementation.


ℹ️ **Repo Maintainers**: To accept this application, [review their application](https://www.drips.network/wave/maintainers/issues/63ff8286-9b72-4dff-aec8-3fec2e682417) or assign @Kureszn to this issue.

usmanimamu17-create:
Closed — addressed by merged PR. 🎉

PR #376, opened 2026-07-30
Fix/workflow issues

Kureszn:
All issues are fixed and should pass
Closes #271


usmanimamu17-create (changes requested):
you did not fix any issue chief, do not do that here.

@Kureszn

Kureszn:
Very sorry about that.
Those changes got dropped during an intended rebase and i was not aware. 
Still working on it actively.

PR #393, opened 2026-08-19
docs(config): document custom severity overwrite behavior and lifecycle

### LINKED ISSUE #383: Custom severity registration lacks uniqueness constraint against canonical severities
**Labels / Complexity:** validation · Trivial — 2 pts

### Problem
The `set_custom_severity` function in `src/config.rs` checks if the severity is canonical using `is_canonical_severity()`, returning `SLAError::InvalidSeverity` if it matches. However, the check only prevents registering a custom severity with the same Symbol as a canonical one. It does not prevent registering the same custom severity twice with different configs, which would silently overwrite the previous config without emitting a "config updated" event specific to custom severities.

**Root cause**
```rust
// src/config.rs:111-120
if crate::SLACalculatorContract::is_canonical_severity(&severity) {
    return Err(SLAError::InvalidSeverity);
}
// <-- No check for existing custom severity with same name
let mut custom: Map<Symbol, SLAConfig> = env
    .storage()
    .instance()
    .get(&CUSTOM_CONFIG_KEY)
    .unwrap_or_else(|| Map::new(env));
custom.set(severity.clone(), SLAConfig { ... });  // <-- Silently overwrites
```

### Why this is architecturally hard
1. Allowing overwrite is intentional for config updates, but the event emitted is `EVENT_CONFIG_UPD`, not a distinct "custom_severity_added" vs "custom_severity_updated".
2. The backend cannot distinguish between new registration and update based on the event alone.

### Proposed design
The current behavior (overwrite with same event) is acceptable. This issue is to document the behavior explicitly. Alternatively, add a `custom_severity_updated` flag to the event payload.

### Acceptance criteria
- [ ] The documentation for `set_custom_severity` explicitly states that calling it with an existing severity updates the config
- [ ] A test verifies that updating an existing custom severity emits the same event structure
- [ ] The `docs/severity-compatibility-matrix.md` explains custom severity lifecycle

### Out of scope
Adding a separate update function for custom severities.

### Getting started
Files in scope:
- `apexchainx_calculator/src/config.rs`

Build/test commands:
```bash
cd apexchainx_calculator
cargo test custom_severity
```

Good first files to read: `src/config.rs` (lines 100-130).

mxllv:
Closes #383

### DIFF
--- a/apexchainx_calculator/src/config.rs
+++ b/apexchainx_calculator/src/config.rs
@@ -102,6 +102,12 @@ pub fn get_last_config_update(env: &Env) -> Result<Option<crate::ConfigUpdateInf
 }
 
 /// Registers or updates a custom (non-canonical) severity level.
+///
+/// # Overwrite & Lifecycle Behavior
+/// - If a custom severity with the given symbol is not registered, it is added to `CUSTCFG`.
+/// - If a custom severity with the given symbol already exists, calling `set_custom_severity`
+///   overwrites the existing parameters (`threshold_minutes`, `penalty_per_minute`, `reward_base`) in-place.
+/// - In both cases, a `cfg_upd` (`EVENT_CONFIG_UPD`) event is emitted with the severity symbol and parameters.
 pub fn set_custom_severity(
     env: &Env,
     severity: Symbol,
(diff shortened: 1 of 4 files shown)


usmanimamu17-create (approved):
LGTM

PR #396, opened 2026-08-20
feat(sla): emit dup_input event with stored result on duplicate rejection

### LINKED ISSUE #385: Duplicate outage input error does not include the existing result
**Labels / Complexity:** api, ux · Medium — 4 pts

### Problem
When `calculate_sla` is called with the same `outage_id` and `config_version_hash`, it returns `SLAError::DuplicateOutageInput`. The error does not include the previously computed result, forcing the caller to make a separate `get_latest_by_outage(outage_id)` call to retrieve the result. This doubles the number of contract calls for the common case of replaying a duplicate submission.

**Root cause**
```rust
// src/calculation.rs:54-56
if prev.config_version_hash == config_version_hash {
    if prev.mttr_minutes != mttr_minutes || prev.threshold_minutes != cfg.threshold_minutes {
        return Err(SLAError::DuplicateOutageInput);  // <-- No result attached
    }
    return Ok(prev);
}
```

### Why this is architecturally hard
1. Soroban error types are enums; attaching data to an error variant requires changing the enum definition.
2. The `SLAError` enum is used in many places; adding a field to `DuplicateOutageInput` affects all callers.
3. The alternative is to return `Ok(prev)` for duplicates, but this conflates "new calculation" with "replay" in the return type.

### Proposed design
Change `SLAError::DuplicateOutageInput` to `SLAError::DuplicateOutageInput { existing: SLAResult }` and update the error handling code. Alternatively, define a new error variant `SLAError::DuplicateWithResult(SLAResult)`.

### Downstream impact
Backend's `translate_contract_result` must handle the new error variant. The frontend does not call the contract directly.

### Acceptance criteria
- [ ] `DuplicateOutageInput` error includes the existing `SLAResult`
- [ ] A test verifies that the error contains the correct result data
- [ ] The error schema is documented in `docs/error-responses.md` or similar

### Out of scope
Changing the success path return type.

### Getting started
Files in scope:
- `apexchainx_calculator/src/lib.rs` (SLAError enum)
- `apexchainx_calculator/src/calculation.rs`

Build/test commands:
```bash
cd apexchainx_calculator
cargo test duplicate
```

Good first files to read: `src/lib.rs` (SLAError definition around line 400).

yellowmam77:
## Summary

Closes #385.

When `calculate_sla` is called with the same `outage_id` and unchanged `config_version_hash` but **conflicting inputs**, the contract previously returned a bare `SLAError::DuplicateOutageInput` with no way to recover the already-stored result. Callers had to issue a second `get_latest_by_outage` read.

## Key finding & approach

The issue's proposed design — attaching the `SLAResult` to the error variant — is not implementable in Soroban: `#[contracterror]` enums are **unit-only u32 codes**. The macro generates `match` arms that reference each variant by name with no field binding, and the wire format is a bare code (verified against `soroban-sdk-macros` 21.7.7 source).

The idiomatic, non-breaking fix is to emit a diagnostic **`dup_input` event** carrying the full stored `SLAResult` immediately before returning the error. Because the error is a returned value (not a host trap), the event is persisted in the same transaction's event log, so consumers recover the conflicting result without a follow-up contract read.

## Changes

- **`apexchainx_calculator/src/lib.rs`** — new `EVENT_DUP_INPUT` constant; `calculate_sla` emits the event (via a `publish_duplicate_input_event` helper) on the conflict path; docs updated on the error variant and method.
- **`apexchainx_calculator/src/calculation.rs`** — mirrored in the delegated implementation for consistency.
- **`apexchainx_calculator/src/event_schema.rs`** — registered `dup_input` in the catalog + distinctness test.
- **`apexchainx_calculator/src/api_stability.rs`** — added `dup_input` to the public event ABI list.
- **Tests** — `test_385_conflicting_duplicate_emits_dup_input_with_stored_result` (verifies the event payload matches the stored result) and `test_385_exact_replay_does_not_emit_dup_input` (replay stays side-effect free).
- **Docs** — `docs/FAILURE_TAXONOMY.md`, `docs/RESULT_PAYLOAD_HASHING.md`, `README.md`, `docs/PROJECT_CONTEXT.md`, `CHANGELOG.md`.

## Event schema

`dup_input` — `topic[0]=dup_input`, `topic[1]=v1`, `topic[2]=severity`; payload mirrors the full `SLAResult` field order:

`(outage_id, status, mttr_minutes, threshold_minutes, amount, payment_type, rating, config_version_hash, recorded_at)`

## Verification

```
cd apexchainx_calculator
cargo fmt --check
cargo clippy -- -D warnings
cargo test --lib        # 572 passed
cargo check --target wasm32-unknown-unknown --lib
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
   internal severity list invariant is ever broken the call surfaces a
   deterministic `InvalidSeverity` error rather than an unrecoverable host trap.
 ### Added
+- `dup_input` event — emitted on the `DuplicateOutageInput` rejection path carrying the previously stored `SLAResult`, so consumers can reconcile a conflicting duplicate from the same transaction's event log without a follow-up `get_latest_by_outage` call (closes #385)
 - Per-severity telemetry counter saturation regression coverage — documented the `u32` lane saturation behavior in the `record_severity_telemetry` code docs and `docs/CONTRACT_MAINTENANCE_POLICY.md`, and added `test_severity_telemetry_counters_saturate_at_u32_max` to verify counters saturate at `u32::MAX` instead of wrapping (release) or panicking (debug) (#387)
 - `docs/CONTRACT_SHAPE_CHANGE_CHECKLIST.md` — release-readiness checklist for PRs that touch storage keys, `STORAGE_VERSION`, event topic constants, or event payload fields; cross-referenced from `CONTRIBUTING.md` as SC-100
 - **[SC-509] SLAError Addition Workflow** (#253) — comprehensive contributor guide for adding, deprecating, or reviewing `SLAError` variants without breaking backend adapter logic. See `docs/sla-error-additions-guide.md`.
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ address may call them.
 
 | Method | What it writes |
 |--------|----------------|
-| `calculate_sla` | Appends a result to history, updates cumulative stats and per-severity telemetry, emits `sla_calc` and `set_int` events. Idempotent on exact replay; rejects conflicting duplicates. |
+| `calculate_sla` | Appends a result to history, updates cumulative stats and per-severity telemetry, emits `sla_calc` and `set_int` events. Idempotent on exact replay; rejects conflicting duplicates with a `dup_input` event carrying the stored result. |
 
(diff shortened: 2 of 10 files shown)


usmanimamu17-create (approved):
LGTM

Issue #225, opened 2026-07-23
Create a repository-wide public API stability scoring or check to help maintainers assess compatibility risk


usmanimamu17-create:
Problem Statement
This work item addresses a concrete maintainability, safety, or operating gap that is visible in the current repository shape: a Rust/Soroban smart contract crate centered on apexchainx_calculator, extensive contract tests, and CI workflows that already exercise build, lint, and deployment surfaces.

Why it matters
The codebase is already mature enough that a small gap in documentation, event compatibility, migration posture, or test coverage can become a high-cost production issue. This issue is meant to reduce that risk with a well-scoped, contributor-friendly improvement.

Technical Context
The current repository architecture is documented in README.md, docs/PROJECT_CONTEXT.md, and the active contract implementation under apexchainx_calculator/src/lib.rs. The public contract behavior is shaped by config, history, governance, pause, telemetry, version negotiation, and event schema modules. The intended fix should remain consistent with those existing patterns.

Expected Outcome
Deliver a clear, maintainable improvement that strengthens architecture, security, reliability, observability, testing, or contributor workflow without weakening the contract’s safety posture.

Acceptance Criteria
- Scope is intentionally small and reviewable.
- The implementation preserves the repository’s current no_std, Soroban, and CI conventions.
- Tests, docs, or automation are updated to match the expected behavior.
- The final change is deterministic, auditable, and safe for future contributors to extend.

Implementation Notes
Prefer a single root-cause change over a broad refactor. Wherever possible, reuse the existing contract module layout and keep the change compatible with the repository’s established event and storage versioning conventions.

Files or modules likely to be affected
- apexchainx_calculator/src/lib.rs
- apexchainx_calculator/src/tests.rs
- apexchainx_calculator/src/history.rs
- apexchainx_calculator/src/config.rs
- .github/workflows/ci.yml
- README.md

Dependencies (if any)
No blocking dependencies beyond the repository’s existing development conventions.

Difficulty level
Medium

Estimated effort
M

Labels
enhancement, api

cyberdocs120:
@cyberdocs120 has applied to work on this issue as part of the **Stellar Wave Program**'s 7th wave.



> pls can you kindly assign this to me


ℹ️ **Repo Maintainers**: To accept this application, [review their application](https://www.drips.network/wave/maintainers/issues/8ea1b31f-763b-467c-b801-6afb26a9f10e) or assign @cyberdocs120 to this issue.

[...]

PR #524, opened 2026-08-27
Fix API-surface drift across four guardrail modules

### LINKED ISSUE #418: get_public_api omits get_contract_info, get_storage_footprint_estimate, and get_rent_estimate: the discovery descriptor drifts from the actual contract surface
---ISSUE---

## Problem

`get_public_api` (`src/lib.rs`) builds a hard-coded `PublicApiDescriptor` from a hand-maintained list of `method(...)` entries. The list contains 58 methods, but three public `#[contractimpl]` methods are missing: `get_contract_info` (the recommended startup handshake, `src/contract_info.rs`), `get_storage_footprint_estimate`, and `get_rent_estimate` (both `src/storage_estimation.rs`).

Consequences:

- **Discovery consumers miss the recommended handshake**: `get_contract_info` is documented as "the recommended startup handshake for backend consumers" and "supersedes `get_contract_metadata()` and `get_version_info()`", yet any backend that discovers the API surface via `get_public_api` (its documented purpose: "validate that the deployed contract matches their expected API surface") never learns the method exists.
- **The descriptor silently tolerates surface drift**: nothing compares the hand-written list to the `#[contractimpl]` block, so every future method added to the contract will be missing from the descriptor until someone edits the list.
- **Two of the three missing methods are the only rent/footprint views**: a backend validating deployment cost assumptions through the descriptor sees no way to query storage cost.

## Root cause

```rust
// src/lib.rs, get_public_api — the list is curated by hand and already stale
methods.push_back(method("get_contract_metadata", false, "none", ""));
// ... no entry for get_contract_info, get_storage_footprint_estimate, get_rent_estimate
```

The descriptor was written when `get_contract_info` and the estimation views were added; the list was not updated.

1. The fix cannot be a one-time list edit: the list is inherently redundant with the `#[contractimpl]` block, so the durable solution is either a compile-time test that enumerates the contract surface (e.g. via the SDK's generated client) and compares it to the descriptor, or a macro that derives the descriptor from the impl block.
2. `PublicApiDescriptor` is `#[contracttype]` and append-only (documented): adding the missing entries changes the response shape of an existing endpoint, which the stability guardrails (`src/api_stability.rs`) do not currently cover for this type.
3. The `event` column is also hand-maintained and already imprecise (`initialize` lists auth "admin" while the method requires both admin and operator authorization; see companion issue).

## Proposed design

Add a test (or CI step) that lists every `pub fn` in the `#[contractimpl]` block and asserts `get_public_api` returns a method entry for each, then backfill the three missing entries. Consider deriving `mutates`/`auth` from the actual implementation rather than re-typing them.

## Acceptance criteria

- [ ] `get_public_api` returns entries for `get_contract_info`, `get_storage_footprint_estimate`, and `get_rent_estimate`.
- [ ] A test fails when a new public contract method is added without a matching descriptor entry.
- [ ] `PublicApiDescriptor` still lists methods alphabetically and remains backward-compatible (append-only).

## Out of scope

Fixing the `auth` mislabels in the descriptor (companion issue) and adding new discovery fields.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`get_public_api`), `apexchainx_calculator/src/contract_info.rs`, `apexchainx_calculator/src/storage_estimation.rs`.

### LINKED ISSUE #419: event_name_symbols returns 16 of 20 event names: the ABI-stability guardrail cannot detect changes to cfg_frz, cfg_unfrz, stats_sat, or migrate_done
## Problem

`src/api_stability.rs` defines the canonical event ABI for the stability score:

```rust
pub fn event_name_symbols() -> [&'static str; 16] {
    [
        "sla_calc", "set_int", "cfg_upd", "paused", "unpause", "op_set", "pruned",
        "pruned_a", "adm_prop", "adm_acc", "adm_can", "adm_ren", "op_prop", "op_acc",
        "op_can", "dup_input",
    ]
}
```

But `src/event_schema.rs` declares 20 event-name constants. Missing from the guardrail list: `cfg_frz`, `cfg_unfrz`, `stats_sat`, and `migrate_done` (the last two are backend-critical — `stats_sat` is the saturation signal for the running stats, `migrate_done` is the migration-completion signal).

Consequences:

- **Renaming a backend-critical event passes the stability check**: `test_225_event_symbols_are_well_known` asserts only on the 16 listed names, so changing `stats_sat` to something else would not fail CI.
- **The stability score under-reports breaking change**: `assess_stability()` returns `StabilityScore::A` while four public event names are outside its audit.
- **The catalog itself is inconsistent**: `event_name_symbols()` includes `dup_input` (added later) but not the freeze or saturation events that shipped in the same era, so the list reflects no deliberate cut-off.

## Root cause

The list was last updated when `dup_input` was added and never re-synced with `event_schema.rs`; nothing links the two files.

1. The right fix is a single source of truth: `event_schema.rs` already holds the canonical constants, so `event_name_symbols()` should be derived from them (or a shared `const` array) rather than duplicated.
2. `assess_stability()` returns a score, not a failure; the acceptance bar for "A" is what the team must decide — either the score becomes exhaustive over the real event catalog, or the score is scoped to a documented subset and the four excluded events get their own guardrail.
3. Any change here touches the release checklist and `docs/EVENT_COMPATIBILITY_POLICY.md`, which instructs backends to treat the event ABI as stable.

## Acceptance criteria

- [ ] `event_name_symbols()` covers all 20 event names in `src/event_schema.rs`.
- [ ] A test fails when a new event constant is added to `event_schema.rs` without a corresponding guardrail entry.
- [ ] `assess_stability()` reflects the full event catalog.

## Out of scope

Per-event payload-shape guardrails (companion issues) and `EVENT_MIGRATE_DONE`'s `&str`-vs-`Symbol` inconsistency.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/api_stability.rs`, `apexchainx_calculator/src/event_schema.rs`.

### LINKED ISSUE #420: canonical_field_counts omits ConfigBundle, AuditState, ContractInfo, HistoryPage, and the public-API types: layout drift in the newest contract types escapes the stability check
## Problem

`canonical_field_counts()` in `src/api_stability.rs` returns counts for 17 `#[contracttype]` structs — the set that existed when the stability check was introduced. Structs added since are absent: `ConfigBundle` (`src/config_bundle.rs`), `AuditState` (`src/audit_state.rs`), `ContractInfo` (`src/contract_info.rs`), `HistoryPage`, `PublicApiMethod`, `PublicApiDescriptor`, `SeverityAliasMapping`, `ContractStateFingerprint`, `VersionNegotiationInfo`, `NegotiationOutcome`, `VersionMismatchDetail`, `VersionNegotiationResult`, `HistoryRetentionMetrics`, and `CompensationAction`.

Consequences:

- **The guardrail cannot detect layout drift in the newest ABI**: `test_225_field_counts_match_declared_types` only checks the 17 listed structs, so adding or removing a field on `ContractInfo` or `AuditState` changes serialization for backend consumers with no stability signal.
- **The type whose schema version is explicitly managed — `ContractInfo.schema_version` — is the one most exposed**: the whole point of `CONTRACT_INFO_SCHEMA_VERSION` is that backends detect field changes via the schema version, but the stability score gives maintainers no independent tripwire.
- **The count of "17" itself is asserted**: `assess_stability()` returns `C` when `canonical_field_counts().len() != 17`, so any addition to the guardrail requires editing three places (`canonical_field_counts`, its test, and the length check).

## Root cause

The list was frozen at the 17 structs present at #225 and never extended as new contract types were introduced; the length assertions make extension an explicit chore that no one performed.

1. Making the guardrail exhaustive means deciding whether to include every `#[contracttype]` (including test-only and negotiation types) or to maintain an explicit allowlist with a compile-time check that the allowlist equals the full set.
2. `canonical_field_counts` returns `[(&str, u32); 17]` — a fixed-size array; extending it to N requires changing the return type or switching to a `Vec`/`&[(&str, u32)]`, rippling through `assess_stability`.
3. There is no tooling that lists `#[contracttype]` structs at compile time, so the "all structs" invariant must be enforced by a test that greps source or by disciplined updates — the same failure mode as the event-name catalog (companion issue).

## Acceptance criteria

- [ ] `canonical_field_counts()` covers every public `#[contracttype]` struct in the crate.
- [ ] A test fails when a `#[contracttype]` struct gains or loses a field without a guardrail update.
- [ ] `assess_stability()` remains `A` for the current codebase with the extended list.

## Out of scope

Automating extraction via proc macros and per-field type checks.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/api_stability.rs`, `apexchainx_calculator/src/contract_info.rs`.

### LINKED ISSUE #421: storage_version.rs is an orphan module: is_migration_complete always returns false and a duplicate STORAGE_VERSION_KEY aliases lib.rs's "VER"
## Problem

`src/storage_version.rs` is never declared as a module: `src/lib.rs`'s module list has `pub mod storage_estimation` but no `pub mod storage_version`, and no other file references it. The file is not compiled into the crate at all. Its contents:

```rust
const STORAGE_VERSION_KEY: Symbol = symbol_short!("VER");      // duplicate of lib.rs's key
const MIGRATION_FLAG_KEY: Symbol = symbol_short!("MIGRATED");

pub fn is_migration_complete(env: &Env) -> bool {
    env.storage().instance().get::<Symbol, bool>(&MIGRATION_FLAG_KEY).unwrap_or(false)
}
```

Consequences:

- **A documented migration helper is dead**: `is_migration_complete()` claims to report "whether a storage migration has been completed", but nothing in the crate ever writes `MIGRATED` (the only writer would be `migrate()` in `src/lib.rs`, which stamps `VER` and never touches `MIGRATED`), so even if the module were compiled it would always return `false`.
- **The duplicate `STORAGE_VERSION_KEY` invites aliasing**: lib.rs's own maintenance note says "No two storage keys may share the same Symbol value", yet this file declares the same `"VER"` key a second time; the collision regression test only checks lib.rs's constants, so the duplicate is invisible to it.
- **The module's existence misleads**: maintainers and consumers grep for migration helpers, find this file, and may wire into a function that can never report `true`.

## Root cause

The file was written as part of the migration work but never wired into the crate (`mod storage_version;` is missing) and never re-synced with the actual migration implementation, which uses a different flag scheme.

1. Two options: delete the orphan (and its misleading API) or actually implement `is_migration_complete` — but the real migration path (`migrate()` in lib.rs) has no completion flag and would need a storage-schema change (a new key) plus a bump of `STORAGE_VERSION` to v2, which cascades into `migrate()` arms and migration tests.
2. Orphan `.rs` files are invisible to `cargo machete`/`udeps` (they are not compiled), so no existing CI gate flags this class of file; preventing recurrence needs a lint that lists undeclared `.rs` files (companion issue covers the other orphans).
3. Any fix must keep `migrate()` idempotent and the v0→v1 arm auditable; introducing a flag key means updating `init_missing_storage_defaults` and the migration tests.

## Acceptance criteria

- [ ] `storage_version.rs` is either removed, or declared and made consistent: `is_migration_complete()` reflects the actual migration state.
- [ ] No two storage-key constants share a Symbol value anywhere in the crate.
- [ ] The storage-key collision regression test covers all declared keys including any new flag.

## Out of scope

The other orphan modules (`deployment_policy.rs`, `payload_optimizer.rs`, `policy.rs`, `threshold_config.rs`, `scratch_test.rs`) — see companion issues.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/storage_version.rs`, `apexchainx_calculator/src/lib.rs` (`migrate`, module list, storage-key maintenance note).

BigBen-7:
## Summary

Closes four related "guardrail drifted from the actual API surface" issues in `apexchainx_calculator`:

- Closes #421 — `storage_version.rs` was never declared as a module (dead code) and its duplicate `STORAGE_VERSION_KEY` aliased `lib.rs`'s `"VER"` key. It's now wired in via `pub mod storage_version;`, the duplicate key constant is gone (it reuses `crate::STORAGE_VERSION_KEY`/`STORAGE_VERSION`), and `is_migration_complete()` now genuinely reports whether on-chain storage matches the version this binary expects.
- Closes #420 — `canonical_field_counts()` only covered the 17 struct types present at #225. Added the 14 types introduced since: `ConfigBundle`, `AuditState`, `ContractInfo`, `HistoryPage`, `PublicApiMethod`, `PublicApiDescriptor`, `SeverityAliasMapping`, `ContractStateFingerprint`, `VersionNegotiationInfo`, `NegotiationOutcome` (tracked by variant count, being a fieldless enum), `VersionMismatchDetail`, `VersionNegotiationResult`, `HistoryRetentionMetrics`, `CompensationAction`. `assess_stability()`'s expected count moved from 17 to 31, with dedicated regression assertions added for `ContractInfo` and `AuditState` (the two called out in the issue as most exposed).
- Closes #419 — `event_name_symbols()` only covered 16 of the 20 event names declared in `event_schema.rs`, missing `cfg_frz`, `cfg_unfrz`, `stats_sat`, and `migrate_done` (the last two backend-critical). Added all four; `assess_stability()`'s expected count moved from 16 to 20, and a new test cross-checks the guardrail list against `event_schema`'s actual constants so future additions there get caught.
- Closes #418 — `get_public_api()`'s hand-maintained method list was missing `get_contract_info` (the recommended startup handshake), `get_storage_footprint_estimate`, and `get_rent_estimate`. Added all three in alphabetical order; method count updated from 58 to 61.

## Test plan

- [ ] `just test` (or `cargo test -p apexchainx_calculator`) passes, including the new/updated tests in `api_stability.rs`, `storage_version.rs`, and `tests.rs`.
- [ ] `assess_stability()` still returns `A` for the current codebase.


usmanimamu17-create (approved):
LGTM

PR #533, opened 2026-08-28
fix: resolve snapshot versioning, view replay parity, config bundle hash, and single-address init (#511, #512, #513, #514)

### LINKED ISSUE #511: SLAConfigSnapshot.version is hardcoded "v1" and never bumps: snapshot schema drift is invisible to consumers
## Problem

Both snapshot endpoints return a version label that never changes:

```rust
// get_config_snapshot and get_custom_config_snapshot (src/lib.rs, src/config.rs)
Ok(SLAConfigSnapshot { version: symbol_short!("v1"), entries })
```

The struct doc says `version` is a "Schema version label (e.g., \"v1\")" — a version that is supposed to change when the snapshot's structure changes. It is a literal in every call site.

Consequences:

- **Structural additions ship with no signal**: if `SLAConfigSnapshot` gains a field (e.g. custom entries folded in, per the companion issue), the `version` label still reads "v1", so consumers that version-check snapshots before decoding cannot detect the change — the label's entire purpose is defeated.
- **The label and the actual schema already disagree in meaning**: "v1" now means "canonical-only snapshot"; a future "v2" would be ambiguous (structure change vs. content change), because the label is defined by neither the struct layout nor the content class.
- **There is no test tying the label to anything**: no test asserts "version bumps when entries shape changes" (impossible to write meaningfully today since the label is a constant).

## Root cause

The `version` field was added as a placeholder label and never wired to a versioning mechanism (unlike `RESULT_SCHEMA_VERSION`, which has a constant and sentinel tests).

1. Making the label meaningful requires a schema-version constant (like `RESULT_SCHEMA_VERSION`) bumped deliberately when the snapshot shape changes, plus a sentinel test — the `SLAResult` pattern the repo already uses.
2. Alternatively the field can be removed from `SLAConfigSnapshot` (an ABI change — the struct is `#[contracttype]` and stability-guarded) if consumers never use it.
3. The custom-snapshot companion issue (adding custom entries to snapshots) is exactly the structural change that would exercise the versioning; the two should be planned together so the bump has a real trigger.

## Acceptance criteria

- [ ] The snapshot version label reflects the snapshot's structure (bumped on shape change, with a sentinel test).
- [ ] The versioning rule is documented alongside `RESULT_SCHEMA_VERSION`.
- [ ] Consumers can detect snapshot schema changes.

## Out of scope

Folding custom entries into the snapshot (companion issue) and the `get_result_schema` versioning (which already has `schema_version`).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`get_config_snapshot`, `get_custom_config_snapshot`), `apexchainx_calculator/src/lib.rs` (`RESULT_SCHEMA_VERSION`).

### LINKED ISSUE #512: calculate_sla_view can disagree with calculate_sla on duplicates: the audit path returns a result for inputs the mutating path would reject as DuplicateOutageInput
## Problem

`calculate_sla_view` (`src/lib.rs`) shares `compute_result` with the mutating path but skips the duplicate-detection state machine:

```rust
pub fn calculate_sla_view(env, outage_id, severity, mttr_minutes) -> Result<SLAResult, SLAError> {
    Self::check_version(&env)?;
    let cfg = Self::load_config(&env, &severity)?;
    let config_version_hash = Self::compute_config_version_hash(&env)?;
    Self::compute_result(outage_id, mttr_minutes, &cfg, config_version_hash, env.ledger().timestamp())
}
```

The mutating path would return `DuplicateOutageInput` for the same `(outage_id, mttr)` under an unchanged hash with a conflicting prior entry; the view returns a *new* result with no awareness of the stored entry.

Consequences:

- **The audit view certifies submissions the mutating path rejects**: a backend using `calculate_sla_view` to pre-validate a submission (a natural use of the "view-only" endpoint) sees a valid result; the actual `calculate_sla` then fails with `DuplicateOutageInput` — the pre-validation gave false confidence.
- **The two paths encode different semantics for the same inputs**: the doc says the view "recalculates SLA deterministically" and shares the computation; but the *decision* semantics (replay vs. conflict vs. new generation) exist only in the mutating path, so the view is not a preview of the mutating path.
- **The replay case diverges too**: for a genuinely idempotent resubmission, `calculate_sla` returns the *stored* result (with its original `recorded_at`), while the view computes a fresh result with the current timestamp — even in the agreeing case the outputs differ.

## Root cause

The view was implemented as "pure computation" without the duplicate/replay policy; the docs describe it as an audit path without defining its relationship to the state machine.

1. Making the view agree requires it to read history and apply the duplicate/replay/conflict logic read-only — turning it into a non-trivial duplicate of `calculate_sla`'s Phase 3 (the very duplication the codebase already struggles with — companion issue), or extracting the policy into a shared pure function both paths call.
2. Alternatively the view's contract can be redefined as "pure math, ignoring history" and the docs corrected — but then its value as a pre-validation tool is explicitly off-limits, and consumers must know not to use it that way.
3. Any change must keep `calculate_sla_view` read-only (no storage writes, no events — a documented property) while adding the history read.

## Acceptance criteria

- [ ] `calculate_sla_view` either applies the duplicate/replay policy (read-only) or is documented as math-only with the divergence explicit.
- [ ] A test covers the same inputs through both paths (conflict case and replay case).
- [ ] The view remains side-effect-free.

## Out of scope

The replay-uses-current-config defect (companion issue in batch 02) and the computation duplication (companion issue in batch 01).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`calculate_sla_view`, `calculate_sla`), `apexchainx_calculator/src/calculation.rs`.

### LINKED ISSUE #513: ConfigBundle omits config_version_hash: bootstrap consumers must issue a second RPC for the hash the duplicate-detection policy keys on
## Problem

`ConfigBundle` (`src/config_bundle.rs`) — the "recommended way for backend consumers to bootstrap their configuration cache in one RPC" — contains the snapshot and the result schema but not the config version hash:

```rust
pub struct ConfigBundle {
    pub snapshot: SLAConfigSnapshot,
    pub schema: SLAResultSchema,
}
```

The hash (`get_config_version_hash`) is the key consumers need for duplicate detection (SC-W5-046: replay/conflict/generation decisions all compare `config_version_hash`) and for cache invalidation (#4, `get_last_config_update`).

Consequences:

- **The one-RPC bootstrap misses the one value the policy keys on**: a backend that bootstraps from the bundle and then receives `sla_calc`/`dup_input` events carrying `config_version_hash` (set_int and dup_input carry it) has no baseline hash to compare against — it must issue `get_config_version_hash` separately, defeating the "one RPC" promise.
- **The bundle's snapshot and the hash can disagree by construction**: the bundle is a point-in-time read of config; a separate hash call in a later ledger could reflect a config change — the consumer cannot tell whether the mismatch is a real change or read skew.
- **`get_full_audit_state` has the same gap**: the audit envelope also omits the hash, so the two "one-shot" reads are both missing the single most policy-relevant config value.

## Root cause

`ConfigBundle` was assembled from the two endpoints that existed (`get_config_snapshot` + `get_result_schema`); `get_config_version_hash` was not included when the bundle was designed.

1. Adding `config_version_hash: u64` to `ConfigBundle` is an additive `#[contracttype]` field change (append-only safe per the stability policy) but requires the `canonical_field_counts` guardrail update (companion issue) and consumer regeneration.
2. The hash must be computed from the *same* snapshot read to guarantee bundle-internal consistency — the implementation must compute the hash over the snapshot it returns (or read the hash and snapshot in one call), which the current composition does not guarantee.
3. The audit envelope (companion issue) should get the same field, so the fix should define the canonical "config posture" payload shared by both.

## Acceptance criteria

- [ ] `ConfigBundle` includes `config_version_hash` consistent with its snapshot.
- [ ] A consumer bootstrapping from the bundle can run the duplicate-detection policy without another RPC.
- [ ] The stability guardrail and consumers are updated for the new field.

## Out of scope

The custom-severity snapshot omission (companion issue in batch 06) and the audit-envelope redundancy (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/config_bundle.rs`, `apexchainx_calculator/src/lib.rs` (`get_config_bundle`, `get_config_version_hash`).

### LINKED ISSUE #514: initialize accepts admin == operator with no validation or warning: the two-party consent at deployment can be a single address
## Problem

`initialize` (`src/lib.rs`) requires both addresses to authorize but never requires them to be distinct:

```rust
    ...
    admin.require_auth();
    operator.require_auth();   // two signatures required...
    env.storage().instance().set(&OPERATOR_KEY, &operator);   // ...but they may be the same address
    ...
}
```

The two-party requirement is real (both must sign), yet the same address can fill both roles — one signature satisfies `require_auth` twice.

Consequences:

- **The separation-of-duties the design implies is optional at deployment**: the role model (#28) exists so config/roles (admin) and calculations (operator) are distinct; `initialize` lets a single address hold both, and the runtime setters can recreate that state (companion issue), so the invariant is never established.
- **The two-party consent is illusory in the single-address case**: `require_auth()` on the same address twice is one signature; a deployer reading the code as "two parties must agree" is misled.
- **No signal marks the merged deployment**: `get_full_audit_state` returns admin == operator with no flag, so downstream tooling cannot distinguish "deliberate single-key deployment" from "misconfiguration".

## Root cause

`initialize` was written to require auth from both arguments without considering that the arguments may be equal; no role-distinctness invariant was defined (the setters companion issue shows the same gap).

1. Enforcing distinct roles at `initialize` is a behavior change for any existing deployment that merged them; the error choice (reuse `InvalidInput` vs. new variant) is an ABI decision (companion error-catalog issue).
2. The policy question — must admin and operator always differ? — affects `set_operator`/`propose_operator` (companion issue) and the renounce flow; the invariant should be defined once and enforced at every mutation point.
3. The two-party semantics (both must authorize) and the distinctness rule are separate contracts; the docs should state both, and tests should cover the single-address case.

## Acceptance criteria

- [ ] `initialize` with `admin == operator` is either rejected or explicitly documented as a supported mode.
- [ ] The role-distinctness invariant is documented and enforced consistently across `initialize` and the role setters.
- [ ] Tests cover the equal-address deployment case.

## Out of scope

The role-setter equal-role gap (companion issue in batch 08) and the two-party `initialize` descriptor mislabel (companion issue in batch 03).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`initialize`, `require_admin`, `require_operator`).

Fatimasanusi:
## Summary
Resolves four technical debt issues (#511, #512, #513, and #514) in `apexchainx_calculator` to improve SLA snapshot schema tracking, calculation view consistency, single-RPC backend bootstrapping, and role initialization documentation.

Closes #511
Closes #512
Closes #513
Closes #514

---

### 1. `SLAConfigSnapshot` Versioning (#511)
- Introduced `CONFIG_SNAPSHOT_SCHEMA_VERSION` (`"v1"`) and `CONFIG_SNAPSHOT_SCHEMA_FIELD_COUNT` (`2`) constants in `lib.rs`.
- Replaced hardcoded version literals across canonical and custom snapshot builder methods.
- Added a CI-backed sentinel test `test_config_snapshot_schema_field_count_sentinel` in `schema_migration_tests.rs`.

### 2. Read-Only Duplicate & Replay Policy for `calculate_sla_view` (#512)
- Refactored `calculate_sla_view` (in `calculation.rs` and `lib.rs`) to inspect storage history read-only before computing math.
- Returns `Err(SLAError::DuplicateOutageInput)` on conflicting duplicate entries and replays the stored result on matching inputs.
- Ensures 100% side-effect-free execution (no storage state writes, no event emissions).

### 3. `ConfigBundle` Hash Bootstrapping (#513)
- Added `pub config_version_hash: u64` to `ConfigBundle` struct.
- Updated `get_config_bundle()` to compute and include the configuration hash.
- Updated API stability guardrail field count for `ConfigBundle` from `2` to `3` in `api_stability.rs`.

### 4. Single-Address Deployment Documentation & Test Coverage (#514)
- Documented single-address / merged-role deployment mode (`admin == operator`) in `initialize()` doc comments.
- Added unit test `test_initialize_single_address_merged_roles` covering single-key initialization and subsequent role invocations.

---

## Verification Plan
- [x] Unit tests added/updated for single-address init, `ConfigBundle` hash integrity, view replay/duplicate conflict handling, and snapshot versioning.
- [x] ABI stability field count assertions pass in `api_stability.rs`.

### DIFF
--- a/apexchainx_calculator/src/api_stability.rs
+++ b/apexchainx_calculator/src/api_stability.rs
@@ -89,7 +89,7 @@ pub fn canonical_field_counts() -> [(&'static str, u32); 31] {
         ("FailureCode", 3),
         ("FailureSchema", 2),
         ("HealthcheckResult", 3),
-        ("ConfigBundle", 2),
+        ("ConfigBundle", 3),
         ("AuditState", 10),
         ("ContractInfo", 11),
         ("HistoryPage", 3),
--- a/apexchainx_calculator/src/calculation.rs
+++ b/apexchainx_calculator/src/calculation.rs
@@ -158,6 +158,29 @@ pub fn calculate_sla_view(
     crate::SLACalculatorContract::check_version(env)?;
     let cfg = crate::SLACalculatorContract::load_config(env, &severity)?;
     let config_version_hash = crate::SLACalculatorContract::compute_config_version_hash(env)?;
+
+    let history: Vec<SLAResult> = env
+        .storage()
+        .instance()
+        .get(&HISTORY_KEY)
+        .unwrap_or_else(|| Vec::new(env));
+
+    let mut existing: Option<SLAResult> = None;
+    for i in 0..history.len() {
+        let entry = history.get(i).unwrap();
+        if entry.outage_id == outage_id {
+            existing = Some(entry);
+        }
+    }
+    if let Some(prev) = existing {
+        if prev.config_version_hash == config_version_hash {
+            if prev.mttr_minutes != mttr_minutes || prev.threshold_minutes != cfg.threshold_minutes {
+                return Err(SLAError::DuplicateOutageInput);
+            }
+            return Ok(prev);
+        }
+    }
+
     compute_result(
         outage_id,
         mttr_minutes,
--- a/apexchainx_calculator/src/config.rs
+++ b/apexchainx_calculator/src/config.rs
@@ -8,7 +8,7 @@ use soroban_sdk::{symbol_short, Env, Map, Symbol, Vec};
 
 use crate::{
     config_freeze, config_metadata, SLAConfig, SLAConfigEntry, SLAConfigSnapshot, SLAError, CONFIG_KEY,
-    CUSTOM_CONFIG_KEY, EVENT_CONFIG_UPD, EVENT_VERSION,
+    CONFIG_SNAPSHOT_SCHEMA_VERSION, CUSTOM_CONFIG_KEY, EVENT_CONFIG_UPD, EVENT_VERSION,
 };
 
 /// Sets the SLA configuration for a given severity level.
@@ -75,7 +75,7 @@ pub fn get_config_snapshot(env: &Env) -> Result<SLAConfigSnapshot, SLAError> {
     }
 
     Ok(SLAConfigSnapshot {
-        version: symbol_short!("v1"),
+        version: CONFIG_SNAPSHOT_SCHEMA_VERSION,
         entries,
     })
 }
@@ -201,7 +201,7 @@ pub fn get_custom_config_snapshot(env: &Env) -> Result<SLAConfigSnapshot, SLAErr
     }
 
     Ok(SLAConfigSnapshot {
-        version: symbol_short!("v1"),
+        version: CONFIG_SNAPSHOT_SCHEMA_VERSION,
         entries,
     })
 }
--- a/apexchainx_calculator/src/config_bundle.rs
+++ b/apexchainx_calculator/src/config_bundle.rs
@@ -46,6 +46,8 @@ pub struct ConfigBundle {
     pub snapshot: SLAConfigSnapshot,
     /// Result schema descriptor with symbol mappings.
     pub schema: SLAResultSchema,
+    /// Config version hash corresponding to the snapshot for duplicate detection.
+    pub config_version_hash: u64,
 }
 
 #[cfg(test)]
@@ -72,6 +74,12 @@ mod tests {
         assert!(
             bundle.is_some(),
             "ConfigBundle must be available after initialize()",
+        );
+        let b = bundle.unwrap();
+        assert_eq!(
+            b.config_version_hash,
+            client.get_config_version_hash(),
+            "ConfigBundle config_version_hash must match get_config_version_hash()",
         );
     }
 
@@ -126,6 +134,11 @@ mod tests {
         assert_eq!(entry.config.threshold_minutes, 42);
         assert_eq!(entry.config.penalty_per_minute, 111);
         assert_eq!(entry.config.reward_base, 999);
+        assert_eq!(
+            bundle.config_version_hash,
+            client.get_config_version_hash(),
+            "Bundle config_version_hash must update when config changes",
+        );
     }
 
     #[test]
@@ -144,6 +157,7 @@ mod tests {
 
         assert_eq!(a.snapshot, b.snapshot);
         assert_eq!(a.schema, b.schema);
+        assert_eq!(a.config_version_hash, b.config_version_hash);
         assert_eq!(a.snapshot.entries.len(), 4);
         assert_eq!(a.schema.status_met, symbol_short!("met"));
     }
--- a/apexchainx_calculator/src/lib.rs
+++ b/apexchainx_calculator/src/lib.rs
@@ -177,6 +177,19 @@ pub(crate) const RESULT_SCHEMA_VERSION: u32 = 1;
 /// 6. See `docs/result-schema-migration-guard.md` for the full process.
 pub(crate) const RESULT_SCHEMA_FIELD_COUNT: u32 = 9;
 
+/// Version label of the SLAConfigSnapshot schema exposed via get_config_snapshot().
+/// Incremented/bumped when snapshot layout changes in a breaking way.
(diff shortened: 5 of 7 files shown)


usmanimamu17-create (approved):
Solid fixes, let's get this in.

PR #534, opened 2026-08-28
test: expand test coverage for telemetry, multi-generation history.

### LINKED ISSUE #507: get_full_audit_state performs about ten redundant version checks and re-reads the same instance storage per delegated getter
## Problem

`get_full_audit_state` (`src/lib.rs`) calls `check_version` once, then delegates to nine methods that each call it again and re-read the same instance storage:

```rust
pub fn get_full_audit_state(env: Env) -> Result<AuditState, SLAError> {
    Self::check_version(&env)?;
    let admin = Self::get_admin(env.clone())?;             // check_version + storage read
    let operator = Self::get_operator(env.clone())?;       // check_version + storage read
    let pending_admin = Self::get_pending_admin(env.clone())?;
    let pending_operator = Self::get_pending_operator(env.clone())?;
    let paused = Self::is_paused(env.clone())?;
    let pause_info = Self::get_pause_info(env.clone())?;
    let config_snapshot = Self::get_config_snapshot(env.clone())?;   // 4 more config map reads
    let stats = Self::get_stats(env.clone())?;
    let result_schema = Self::get_result_schema(env.clone())?;
    ...
}
```

Soroban instance storage is read whole per key access, so this "one-shot bootstrap" issues ~10 version checks and many full-key deserializations for data that could be read once.

Consequences:

- **The bootstrap envelope is not actually one-shot-cheap**: the envelope exists so backends avoid N round-trips; internally it performs N reads + N version checks, so the RPC is cheap but the host work is the sum of all delegated reads.
- **The version check is 10x redundant**: `check_version` is pure (reads `VER`); doing it ten times in one call is wasted work with no correctness benefit.
- **The cost grows with each added field**: extending `AuditState` with another delegated getter adds another `env.clone()` + read + version check; nothing measures the envelope's cost (no read-budget test covers `get_full_audit_state`).

## Root cause

The envelope was composed from existing getters rather than written as a single storage pass; the delegation pattern (clone env, delegate, check_version inside) was reused without considering the redundancy.

1. Rewriting the envelope as one storage pass requires reading each key directly (or adding internal helpers that skip `check_version` — e.g. `pub(crate)` unchecked getters) — a refactor that must preserve the exact response semantics (including `pause_info` as `Vec`, the `history_len` full read — companion issues).
2. The `env.clone()` pattern is pervasive; the fix should establish whether internal composition should use unchecked helpers vs. public methods.
3. Any read-budget enforcement (the `offchain/readCostRegression.ts` companion issue) should cover this envelope, so the fix should produce a measurable cost model.

## Acceptance criteria

- [ ] `get_full_audit_state` performs a single version check and reads each storage key once.
- [ ] Response semantics are unchanged (all existing `AuditState` tests pass).
- [ ] The envelope's read cost is documented (or measured by a read-budget check).

## Out of scope

The `history_len` full-history read (companion issue) and the `pause_info` representation (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`get_full_audit_state`), `apexchainx_calculator/src/audit_state.rs`.

### LINKED ISSUE #508: The telemetry weekly reset wipes fresh violations: calc_stale OR violation_stale resets both lanes, losing a violation recorded a day ago
## Problem

`record_severity_telemetry` (`src/lib.rs`/`src/calculation.rs`) resets both counters when either lane is stale:

```rust
let calc_stale = last_calc != 0 && now.saturating_sub(last_calc) >= week_seconds;
let violation_stale = last_violation != 0 && now.saturating_sub(last_violation) >= week_seconds;
if calc_stale || violation_stale {
    calculations = set_count_lane(calculations, index, 0);
    violations = set_count_lane(violations, index, 0);
}
```

Consider: the last calculation for a severity was 8 days ago (stale), but the last violation was 1 day ago (fresh, `violations = 5`). A new calculation today triggers the `calc_stale` branch and resets the violation counter to 0 — the 5 fresh violations from yesterday are erased.

Consequences:

- **The weekly window under-counts real violations**: the violation rate for the current week is computed after the wipe, so a severity that violated 5 times this week loses all of them because its *calculation* lane was stale — the two metrics are incorrectly coupled.
- **The docstring's "Lane Isolation" claim is violated**: `calculation.rs`'s docs promise "Reset is per-severity lane. Calculations or inactivity in one severity level do not reset telemetry for other severities" — but within a severity, calculation staleness resets the *violation* lane, coupling the two counters the docs treat as independent.
- **The reset is asymmetric**: a fresh calculation with a stale violation lane correctly resets both (the violation data is genuinely old), but the reverse coupling (stale calc wiping fresh violations) has no justification.

## Root cause

The reset condition was written as `calc_stale || violation_stale` for both counters, treating staleness as a lane-level property instead of a per-counter property.

1. The correct semantics are per-counter: reset `calculations` when `calc_stale`, reset `violations` when `violation_stale` — but the counters share the packed `u128` lanes, so the fix must set/reset lanes independently (the existing `set_count_lane` supports this).
2. The weekly-window design (7 days, lazy reset on next call) is documented; the fix must preserve the lazy-reset contract while decoupling the two lanes, and the doc's "Lane Isolation" wording should be updated to describe per-counter isolation.
3. Tests must pin the coupling case: stale calc + fresh violation → violation counter survives. No such test exists today (the telemetry tests assert the simple reset case).

## Acceptance criteria

- [ ] A fresh violation lane survives a stale-calculation reset.
- [ ] Each counter resets only on its own staleness.
- [ ] The docs describe per-counter isolation, and tests cover the stale-calc/fresh-violation case.

## Out of scope

The u32 timestamp clamp (companion issue) and custom-severity lane misattribution (companion issue in batch 01).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`record_severity_telemetry`), `apexchainx_calculator/src/calculation.rs` (same function).

### LINKED ISSUE #509: get_history_by_outage returns entries across config generations with no distinguishing signal: consumers cannot isolate the current decision
## Problem

`get_history_by_outage` (`src/lib.rs`/`src/history.rs`) returns every retained entry for an outage:

```rust
for i in 0..history.len() {
    let entry = history.get(i).unwrap();
    if entry.outage_id == outage_id { matches.push_back(entry); }
}
```

With the duplicate-detection policy (SC-W5-046), an outage can legitimately have multiple entries — one per config generation, up to `MAX_RECALCS_PER_OUTAGE` (16). Entries differ by `config_version_hash`, but the endpoint returns them flat with no field or ordering note marking which is the current generation.

Consequences:

- **Consumers must re-derive the generation logic**: to answer "what is the current decision for this outage?", a caller must compare `config_version_hash` against `get_config_version_hash` (a second RPC) and pick the matching entry — the endpoint that exists to serve per-outage history makes the common case (latest decision) require client-side logic.
- **The `get_latest_by_outage` sibling exists precisely because this is hard**: the two endpoints split the work (all vs. latest), but a consumer that called `get_history_by_outage` (e.g. to see recalcs) has no per-entry marker (generation index, hash-vs-current flag) to filter on.
- **The event stream has the same gap**: `sla_calc` events don't carry `config_version_hash` (companion issue), so even event-replaying consumers cannot attribute entries to generations.

## Root cause

The endpoint was written as a simple filter when the duplicate/recalc policy was simpler; the generation semantics arrived later (#93-era and SC-W5-046) without extending the accessor.

1. Enriching the response (e.g. a per-entry `is_current` flag or returning entries grouped by generation) changes the endpoint's return type (`Vec<SLAResult>` → a wrapper struct) — an ABI change to a public read endpoint, stability-guarded.
2. "Current generation" is relative to the *current* config hash; the flag would be computed at read time and could go stale within the response window if the config changes in the same ledger — the semantics must be pinned.
3. The fix overlaps with the `sla_calc`-omits-hash companion issue: if the event carried the hash, event-based consumers could attribute generations without an endpoint change.

## Acceptance criteria

- [ ] A consumer can identify the current-generation entry from `get_history_by_outage` (or a documented equivalent).
- [ ] Multi-generation results are distinguishable (field, order, or grouping).
- [ ] Tests cover a multi-generation outage history.

## Out of scope

The `sla_calc` hash omission (companion issue) and the history storage redesign (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`get_history_by_outage`, `get_latest_by_outage`), `apexchainx_calculator/src/history.rs` (same functions).

### LINKED ISSUE #510: prune_history_by_age silently no-ops for min_age_seconds >= now: the cutoff saturates to 0 and no event fires, so callers cannot tell nothing was pruned
## Problem

`prune_history_by_age` (`src/lib.rs`/`src/history.rs`) saturates the cutoff:

```rust
let now = env.ledger().timestamp();
let cutoff = now.saturating_sub(min_age_seconds);
...
if entry.recorded_at >= cutoff { new_history.push_back(entry); } else { removed += 1; }
...
if removed > 0 { ...; env.events().publish(...); }
```

When `min_age_seconds >= now`, `cutoff == 0`, and since every stored entry has `recorded_at >= 0` (a live timestamp), `removed == 0` — the call succeeds, prunes nothing, and emits no `pruned_a` event. An admin passing an accidentally huge age (e.g. `9_999_999_999`) gets a silent no-op.

Consequences:

- **The no-op is indistinguishable from success**: the `pruned_a` event only fires when `removed > 0`, so "pruned 0 because nothing is old enough" and "pruned 0 because the age argument was nonsense" look identical; the admin cannot tell whether the retention policy is working.
- **The endpoint accepts semantically invalid input**: a `min_age_seconds` larger than the contract's entire lifetime can never prune anything; nothing validates it, and the docstring doesn't state the range.
- **The `recorded_at == 0` carve-out compounds the confusion**: the doc says entries with `recorded_at == 0` are always kept (companion issue notes these cannot occur via the mutating path), so the cutoff math has two silent special cases.

## Root cause

`saturating_sub` was used defensively (avoiding underflow) without a corresponding validation or no-op signal; the endpoint was written to be total (always `Ok`).

1. The fix options: (a) validate `min_age_seconds` against a sane bound (e.g. reject ages that make `cutoff == 0` unless the intent is "prune everything") — an error-surface change; (b) emit the `pruned_a` event even for `removed == 0` (a behavior change consumers may rely on for "nothing to do"); (c) document the saturation semantics precisely.
2. The event-only-on-change convention (shared with `prune_history`, companion issue) means "no event = nothing removed"; making the no-op observable conflicts with that convention unless the docs carve out this endpoint.
3. The `recorded_at == 0` carve-out means the cutoff can also be interpreted as "keep everything pre-timestamp"; the fix should define the interaction.

## Acceptance criteria

- [ ] A `min_age_seconds >= now` call is either rejected or produces an observable signal.
- [ ] The docstring states the valid range and the no-op semantics.
- [ ] Tests cover the cutoff edge (`min_age == now`, `min_age > now`).

## Out of scope

The `recorded_at == 0` doc drift (companion issue in batch 07) and the event-on-no-op convention for `prune_history`.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`prune_history_by_age`), `apexchainx_calculator/src/history.rs` (same function).

Agbelomo:
# Refactor `get_full_audit_state` single-pass storage access, decouple telemetry resets, support multi-generation outage history, and validate `prune_history_by_age`

Closes #507
Closes #508
Closes #509
Closes #510

## Summary

This PR addresses four technical debt items in the `apexchainx_calculator` contract across storage access efficiency, telemetry counter reset logic, outage history query semantics, and age-based retention pruning validation:

1. **#507 - Single-Pass Storage Access for `get_full_audit_state`**: Refactored `get_full_audit_state` in `apexchainx_calculator/src/lib.rs` to execute a single `check_version(&env)` check and read storage keys (`ADMIN_KEY`, `OPERATOR_KEY`, `PENDING_ADMIN_KEY`, `PENDING_OP_KEY`, `PAUSED_KEY`, `PAUSE_INFO_KEY`, `STATS_KEY`, `HISTORY_KEY`) directly in one pass. Optimized `get_config_snapshot` and `build_config_snapshot` to read `CONFIG_KEY` once.
2. **#508 - Decoupled Telemetry Resets**: Decoupled `calc_stale` and `violation_stale` counter reset logic in `record_severity_telemetry` (`calculation.rs` & `lib.rs`). Each lane counter is now evaluated and reset independently, ensuring fresh violation data survives stale calculation resets (and vice versa).
3. **#509 - Multi-Generation Outage History**: Documented and verified `get_history_by_outage` ordering (oldest-first, latest decision last). Each entry carries its `config_version_hash`, enabling consumers to distinguish entries across configuration generations.
4. **#510 - `prune_history_by_age` Range Validation**: Added explicit range validation to `prune_history_by_age` in `history.rs` and `lib.rs`. Invocations with `min_age_seconds >= current_ledger_timestamp` are rejected with `Err(SLAError::InvalidInput)` to prevent silent no-ops.

## Changes Made

- `apexchainx_calculator/src/lib.rs`: Single-pass `get_full_audit_state`, optimized `build_config_snapshot`, decoupled telemetry reset, `min_age_seconds` validation, updated docstrings.
- `apexchainx_calculator/src/calculation.rs`: Decoupled `calc_stale` and `violation_stale` reset logic in `record_severity_telemetry`.
- `apexchainx_calculator/src/history.rs`: Added `min_age_seconds >= now` validation returning `Err(SLAError::InvalidInput)`, updated `get_history_by_outage` docstring.
- `apexchainx_calculator/src/tests.rs`: Added unit tests `test_get_full_audit_state_single_pass_efficiency`, `test_record_severity_telemetry_decoupled_lane_resets`, `test_get_history_by_outage_multi_generation_history`, `test_prune_by_age_rejects_min_age_equal_to_now`, and `test_prune_by_age_rejects_min_age_greater_than_now`.

## Verification

- `cargo check --tests`: **PASSED**
- `cargo test`: **PASSED** (all 594 test cases passing)


usmanimamu17-create (approved):
Test coverage looks great, merging.

usmanimamu17-create (approved):
LGTM, thanks!

PR #536, opened 2026-08-30
test: add migration harness, history page matrix, and behavioral pinn…

### LINKED ISSUE #503: No test pins the behavior of get_history_page and get_history_page_with_meta against each other: the documented equivalence (identical items) is unasserted
## Problem

`get_history_page_with_meta` (`src/lib.rs`/`src/history.rs`) promises: "The `items` slice is identical to what `get_history_page` returns for the same `(offset, limit)`". The two accessors are implemented separately (companion issue on the duplication), and the test suite does not contain a direct comparison test — the pagination tests exercise each accessor's own contract but not the equivalence the docstring promises.

Consequences:

- **The equivalence guarantee is unverified**: if the two implementations diverge (e.g. one clamps `limit` differently — the saturating logic is duplicated), consumers that switch between the two accessors (the meta variant exists so consumers can avoid a second call) get different `items` for the same arguments with no test catching it.
- **The docstring is the only spec**: the "identical items" claim lives in prose; a future refactor of one accessor can silently break it.
- **The `total`/`has_more` fields compound the risk**: `has_more` is computed from the same `end` used for slicing, so a slicing bug corrupts the pagination metadata too — and nothing cross-checks `has_more` against a manual count.

## Root cause

The two accessors were written together but tested separately; the equivalence invariant was documented, not asserted.

1. The direct fix is a parametrized test (for each offset/limit pair, assert `items` equality and `has_more` correctness against a known history) — mechanically simple, but it must be written against the *policy* (issue #263) not the current implementation, so the test survives the storage redesign (companion issue).
2. The equivalence is only meaningful with a defined history state; the test must build a history of known length and iterate a range of offsets/limits including the edge cases (offset >= len, limit 0, limit > remaining).
3. This test is a prerequisite for the de-duplication refactor (companion issue): consolidating the two implementations is only safe if the equivalence test exists first.

## Acceptance criteria

- [ ] A test asserts `get_history_page_with_meta(offset, limit).items == get_history_page(offset, limit)` across a matrix of offsets/limits on a fixed history.
- [ ] The test covers edge cases (empty history, offset beyond end, limit 0, limit exceeding remaining).
- [ ] `has_more` is asserted against a manual count for the same matrix.

## Out of scope

The pagination de-duplication (companion issue) and the storage redesign (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`get_history_page`, `get_history_page_with_meta`), `docs/HISTORY_PAGINATION_POLICY.md`.

### LINKED ISSUE #504: The economic-exposure and telemetry views are not exercised against a contract with custom severities: the canonical-only implementations are tested only on canonical state
## Problem

The tests for `get_economic_exposure` and `get_severity_telemetry` (in `src/tests.rs`) run against contracts with only the four canonical severities. The companion issues document that both views are canonical-only by implementation: `get_economic_exposure` iterates `canonical_severities` (custom configs absent), and `get_severity_telemetry` maps custom-severity activity into the critical lane via `canonical_severity_index(...).unwrap_or(0)`.

Consequences:

- **The custom-severity behavior is untested**: the misattribution (custom activity counted in the critical lane) and the omission (exposure ignores custom configs) are visible only if a test registers a custom severity and asserts the views' output — no such test exists, so the defects shipped without coverage.
- **The views' docs and behavior diverge silently**: both docstrings claim full coverage ("across all configured severities", "per-severity weekly") while the tests certify only the canonical case — the tests actively reinforce the wrong reading.
- **The fix direction is uninformed**: when the companion issues are implemented (include custom severities, fix lane indexing), the change cannot be validated against the pre-fix behavior because no test pins it.

## Root cause

The views predate or ignore #93; their tests were written against the canonical-only state the views were designed for, and never revisited when custom severities became first-class in `calculate_sla`.

1. The tests must first pin the *current* (defective) behavior for custom severities (so the fix's behavior change is explicit), then flip to the corrected assertions — a two-phase change that the issue should sequence.
2. The telemetry lane indexing (custom → critical) and the exposure omission are separate defects with separate fixes (companion issues); the test suite should cover both so they can land independently.
3. Custom-severity test state requires `set_custom_severity` calls with valid general bounds; the tests must pick values that stay inside the cross-parameter consistency rule (penalty*3 < reward*2).

## Acceptance criteria

- [ ] Tests register a custom severity and assert `get_economic_exposure` output (pin current behavior, then corrected behavior).
- [ ] Tests run `calculate_sla` against a custom severity and assert `get_severity_telemetry` output.
- [ ] The tests fail when the canonical-only behavior regresses in either direction.

## Out of scope

The fixes themselves (companion issues in batch 06/batch 01) and the telemetry overflow edge (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`get_economic_exposure`, `get_severity_telemetry`), `apexchainx_calculator/src/tests.rs` (view tests).

### LINKED ISSUE #505: Migration tests cover only v0-to-v1: the migrate() harness's "each arm is never modified" rule has no test that future arms chain correctly
## Problem

`migrate()` (`src/lib.rs`) is designed for multi-step migration:

```rust
// Apply each step in sequence. Each arm must be a pure, atomic transformation:
// read old state → write new state → bump version. A future version bump adds
// a new arm here; existing arms are never modified so older migration paths
// remain auditable.
if current == 0 { ... current = 1; }
// v1 → v2 (placeholder for the next breaking state change):
// if current == 1 { ... current = 2; }
```

The schema-migration tests (`src/schema_migration_tests.rs`) cover the v0→v1 path (the only real arm) and the version-mismatch rejections. There is no test that exercises a *chained* migration (v0 → v1 → v2) because v2 does not exist — but more importantly, there is no test scaffolding that would make the next contributor's arm addition verifiable (the placeholder arm is commented out, so a contributor uncommenting it cannot run a test proving old + new arms chain).

Consequences:

- **The first real v2 migration ships untested for chaining**: when `STORAGE_VERSION` is bumped to 2 and the arm is written, there will be no pre-existing test pattern proving a v0-deployed contract reaches v2 in one `migrate()` call (the doc's "multiple versions behind is brought fully up to date in one call" promise).
- **The placeholder arm is dead code with no test contract**: the commented-out `v1 → v2` block documents intent, but nothing verifies that the arm-sequencing pattern (current tracking, idempotency, `VersionMismatch` on unknown versions) is preserved when it is enabled.
- **The migration checklist (`docs/STORAGE_KEY_MIGRATION_CHECKLIST.md`) has no automated companion**: each new arm must be reviewed manually; a test harness that synthesizes a v1 state and runs `migrate()` would make arm additions mechanical.

## Root cause

The migration harness was built with exactly one arm; the test suite covers the one arm, and the multi-arm chaining behavior is documented but unexercised.

1. Testing future arms requires the ability to construct a v1-state contract (write `VER = 1` plus the v1 key layout) in the Env and run `migrate()` — the scaffolding is straightforward, but it must be written *now*, before v2 exists, so it can be extended when the arm lands.
2. The chaining promise ("multiple versions behind in one call") only becomes testable when two arms exist; the issue should define a synthetic-version test pattern (a test-only arm or a mock step) that proves the loop logic without waiting for a real v2.
3. The arm design ("existing arms are never modified") means each new arm must be appended; a test asserting arm immutability (e.g. a snapshot of the v0→v1 arm's behavior) would enforce the auditability rule mechanically.

## Acceptance criteria

- [ ] A test pattern exists for constructing a past-version contract state and running `migrate()` through multiple arms.
- [ ] The chaining/idempotency logic is covered by a synthetic two-arm test (or the test-only step is documented as the pattern for v2).
- [ ] The migration checklist references the test pattern as the required companion for new arms.

## Out of scope

Adding the v2 migration itself and the storage-layout tests beyond the harness.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`migrate`), `apexchainx_calculator/src/schema_migration_tests.rs`, `docs/STORAGE_KEY_MIGRATION_CHECKLIST.md`.

### LINKED ISSUE #506: ContractInfo.needs_migration is unreachable: get_contract_info returns VersionMismatch before constructing the struct, contradicting its own docs and the startup-handshake purpose
---ISSUE---

## Problem

`get_contract_info` (`src/contract_info.rs`) — the "recommended startup handshake for backend consumers" — calls `check_version` first, which errors out in the exact case the struct's `needs_migration` field exists to report:

```rust
pub fn get_contract_info(env: &Env) -> Result<ContractInfo, SLAError> {
    SLACalculatorContract::check_version(env)?;   // <- returns VersionMismatch when stored != STORAGE_VERSION
    ...
    let needs_migration = stored_version != STORAGE_VERSION;   // <- unreachable true
    ...
}
```

The module's own test proves the field is dead:

```rust
// src/contract_info.rs, test_contract_info_detects_migration_needed
env.storage().instance().set(&crate::STORAGE_VERSION_KEY, &99u32);
let err = get_contract_info(&env).unwrap_err();
assert_eq!(err, crate::SLAError::VersionMismatch);   // the struct is never built
```

Consequences:

- **The recommended handshake cannot be performed in the state it exists to detect**: `docs/MIGRATION_STATE_CONSUMPTION.md` tells backends to check `needs_migration` after an upgrade; calling the *recommended* `get_contract_info` during exactly that state yields `VersionMismatch` — the field never reports `true`.
- **The doc contract is false in both directions**: the struct doc says "True when stored storage version differs from the binary's expected version" (unreachable), and the method doc says it "remains callable"... it does not — it errors pre-migration, while `get_version_info`/`get_migration_state`/`get_contract_state_fingerprint` all deliberately bypass `check_version` for this reason.
- **Backends following the docs get the wrong failure**: a consumer using the documented "check `schema_version`/`needs_migration` before decoding" protocol receives an error instead of a struct, and must special-case the one endpoint that was supposed to unify the version story.

## Root cause

`get_contract_info` was written before the "version endpoints bypass `check_version`" convention was established; when `get_migration_state` et al. were made pre-migration-safe, this endpoint was not.

1. The fix is removing the `check_version` call and relying on the struct's own `needs_migration`/`storage_version` fields (the data is already read without the guard) — but the change must be verified against every test and consumer that expects `VersionMismatch` from `get_contract_info` (the module's own test asserts it).
2. This is a behavior change to the recommended endpoint: consumers that currently rely on catching `VersionMismatch` from `get_contract_info` must switch to reading the struct — a coordinated consumer migration.
3. The endpoint family's version semantics should be reconciled: four endpoints (`get_version_info`, `get_migration_state`, `get_contract_state_fingerprint`, `get_contract_info`) have three different pre-migration behaviors; the fix should document which behavior is canonical (companion issue in batch 03).

## Acceptance criteria

- [ ] `get_contract_info` returns a struct with `needs_migration: true` when stored version != expected.
- [ ] The module's `test_contract_info_detects_migration_needed` asserts the new behavior.
- [ ] The startup-handshake docs match the implemented behavior.

## Out of scope

Consolidating the four version endpoints (companion issue) and the hardcoded `contract_version` (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/contract_info.rs` (`get_contract_info`), `apexchainx_calculator/src/lib.rs` (`check_version`, `get_migration_state`).

Alhaji-naira:
closes #503 

closes #504 

closes #505 

closes #506

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,8 @@ node_modules/
 .env
 .env.*
 !.env.example
+
+
+/target/
+**/*.rs.bk
+*.wasm
--- a/apexchainx_calculator/src/contract_info.rs
+++ b/apexchainx_calculator/src/contract_info.rs
@@ -82,14 +82,15 @@ pub struct ContractInfo {
 /// identity, version posture, and feature availability before resuming
 /// operations. The `schema_version` field lets consumers detect when
 /// new fields have been added to `ContractInfo`.
+///
+/// This function intentionally bypasses `check_version()` so backend consumers
+/// can observe `needs_migration == true` pre-migration during startup handshake.
 pub fn get_contract_info(env: &Env) -> Result<ContractInfo, SLAError> {
-    SLACalculatorContract::check_version(env)?;
-
     let stored_version: u32 = env
         .storage()
         .instance()
         .get(&crate::STORAGE_VERSION_KEY)
-        .unwrap_or(0);
+        .ok_or(SLAError::NotInitialized)?;
     let needs_migration = stored_version != STORAGE_VERSION;
     let is_paused: bool = env.storage().instance().get(&crate::PAUSED_KEY).unwrap_or(false);
 
@@ -114,7 +115,7 @@ pub fn get_contract_info(env: &Env) -> Result<ContractInfo, SLAError> {
         schema_version: CONTRACT_INFO_SCHEMA_VERSION,
         contract_name: symbol_short!("sla_calc"),
         contract_version: symbol_short!("0_1_0"),
-        storage_version: STORAGE_VERSION,
+        storage_version: stored_version,
         result_schema_version: RESULT_SCHEMA_VERSION,
         event_version: crate::event_schema::current_event_version(),
         needs_migration,
@@ -253,13 +254,13 @@ mod tests {
             let info = get_contract_info(&env).unwrap();
             assert!(!info.needs_migration);
 
-            // Corrupt the stored version: get_contract_info must now surface
-            // an explicit VersionMismatch error (the contract reports that it
-            // needs migration instead of returning stale metadata).
+            // Change stored version: get_contract_info returns a ContractInfo
+            // struct with needs_migration: true and the stored storage_version.
             env.storage().instance().set(&crate::STORAGE_VERSION_KEY, &99u32);
 
-            let err = get_contract_info(&env).unwrap_err();
-            assert_eq!(err, crate::SLAError::VersionMismatch);
+            let info_mig = get_contract_info(&env).unwrap();
+            assert!(info_mig.needs_migration);
+            assert_eq!(info_mig.storage_version, 99u32);
         });
     }
 
--- a/apexchainx_calculator/src/schema_migration_tests.rs
+++ b/apexchainx_calculator/src/schema_migration_tests.rs
@@ -216,4 +216,47 @@ mod tests {
             panic!("get_config_bundle returned None after initialization");
         }
     }
+
+    // -----------------------------------------------------------------------
+    // Multi-arm migration harness and chaining test pattern (#505)
+    // -----------------------------------------------------------------------
+
+    /// **#505 – Multi-arm migration chaining & idempotency test pattern.**
+    ///
+    /// Demonstrates and validates the multi-step migration pattern (`v0 -> v1 -> ...`).
+    /// Verifies that a contract starting at a historical storage version (`v0`)
+    /// advances through all sequential migration arms to reach `STORAGE_VERSION`
+    /// in a single `migrate()` call, and that subsequent calls are idempotent no-ops.
+    #[test]
+    fn test_multi_arm_migration_chaining_and_idempotency() {
+        use crate::STORAGE_VERSION;
+        let (env, client) = setup();
+
+        let admin = soroban_sdk::Address::generate(&env);
+
+        // 1. Synthesize a v0 contract state by resetting STORAGE_VERSION_KEY to 0
+        env.storage().instance().set(&crate::STORAGE_VERSION_KEY, &0u32);
+
+        // Verify pre-migration state reports needs_migration = true
+        let mig_state = client.get_migration_state();
+        assert_eq!(mig_state.stored_version, 0);
+        assert_eq!(mig_state.expected_version, STORAGE_VERSION);
+        assert!(mig_state.needs_migration);
+
+        // 2. Invoke migrate() — advances v0 -> v1 (and any subsequent arms sequentially)
+        let result = client.try_migrate(&admin);
+        assert!(result.is_ok(), "migrate() failed on v0 contract state");
+
+        // 3. Verify post-migration state is fully updated to current STORAGE_VERSION
+        let post_state = client.get_migration_state();
+        assert_eq!(post_state.stored_version, STORAGE_VERSION);
+        assert!(!post_state.needs_migration);
+
+        // 4. Idempotency test: calling migrate() again when already current is a safe no-op
+        let retry_result = client.try_migrate(&admin);
+        assert!(retry_result.is_ok(), "second migrate() call must be a safe idempotent no-op");
+        let retry_state = client.get_migration_state();
+        assert_eq!(retry_state.stored_version, STORAGE_VERSION);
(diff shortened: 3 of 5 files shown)


usmanimamu17-create (approved):
Good stuff, approved.

usmanimamu17-create (approved):
Looks good to me 👍

PR #538, opened 2026-08-30
fix(governance): set_operator invalidates a pending operator proposal

### LINKED ISSUE #469: set_operator can orphan a pending operator proposal: the two-step handoff state survives a direct assignment
## Problem

Two operator-change paths coexist in `src/governance.rs`:

```rust
// Two-step: propose_operator -> accept_operator (writes PENDING_OP_KEY)
// Single-step: set_operator (writes OPERATOR_KEY directly, admin only)
pub fn set_operator(env: &Env, caller: &Address, new_operator: &Address) -> Result<(), SLAError> {
    ...
    env.storage().instance().set(&OPERATOR_KEY, new_operator);
    env.events().publish((EVENT_OP_SET, EVENT_VERSION, caller.clone()), (new_operator.clone(),));
    Ok(())
}
```

`set_operator` does not clear `PENDING_OP_KEY`. If a proposal is pending, a direct `set_operator` installs a new operator while the pending slot still names someone else; that someone can later call `accept_operator` and silently replace the operator that was just set directly.

Consequences:

- **A stale proposal can override a direct assignment**: the sequence set_operator(B) → (later) accept_operator(C, pending from before) leaves C as operator, undoing the admin's direct decision — with no event ordering that explains it.
- **The two-step protocol is bypassable and then re-assertable**: `set_operator` exists as a "single-step, admin only" escape hatch, but its failure to clear the pending slot lets the two-step path fire *after* the direct one, producing an operator the admin never intended.
- **Audit trails cannot reconcile the two paths**: `op_set` and `op_acc` events are independent; nothing links a direct assignment to the pending proposal it should have invalidated.

## Root cause

`set_operator` and the proposal lifecycle were written independently; the invariant "a direct assignment invalidates any pending proposal" was never encoded.

1. The fix (clear `PENDING_OP_KEY` in `set_operator`, possibly emitting `op_can`) is small, but it changes event behavior and must be paired with the state-machine definition from the propose-side companion issue.
2. There is a design question whether `set_operator` should even exist alongside the two-step path (the event schema documents both, but only the two-step path has cancel semantics); the issue should decide the canonical path.
3. Tests must cover the interleaving (pending → direct set → accept) and pin the resulting operator and event stream.

## Acceptance criteria

- [ ] A direct `set_operator` invalidates any pending operator proposal (cleared or rejected).
- [ ] The event stream reflects the invalidation.
- [ ] Tests cover pending → direct set → accept interleavings.

## Out of scope

Removing `set_operator` entirely and the admin-side blind set (companion issues).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/governance.rs` (`set_operator`, `propose_operator`, `accept_operator`).

beulah7717108-eng:
## Summary
`set_operator` installed the operator directly (admin-only) but left `PENDING_OP_KEY` intact, so a stale handoff could later complete and silently override the admin's single-step decision — the sequence set_operator(B) then accept_operator(C) would leave C as operator.

This change makes a direct assignment invalidate any pending operator proposal: it clears the pending slot and its expiry timestamp and emits an `op_can` event when one is pending, so the two-step and single-step paths can no longer conflict.

## Changes
- `governance.rs`: `set_operator` now clears `PENDING_OP_KEY` / `PENDING_OP_TS_KEY` and emits `op_can` when a proposal was pending.
- `tests.rs`: coverage for the invalidation on direct set and for the blocked stale-accept path.

## Tests
New tests pins the cleared pending slot, the installed operator, and that the superseded candidate can no longer accept.

Closes #469


usmanimamu17-create (approved):
Good fix, let's merge.

usmanimamu17-create (approved):
Nice fix, merging 👍

PR #539, opened 2026-08-30
feat(governance): emit distinct supersession event on role re-proposal

### LINKED ISSUE #468: propose_admin silently overwrites a pending proposal: the superseded candidate is never notified and no event distinguishes replacement
---ISSUE---

## Problem

`propose_admin` (`src/governance.rs`) writes the pending slot unconditionally:

```rust
pub fn propose_admin(env: &Env, caller: &Address, new_admin: &Address) -> Result<(), SLAError> {
    ...
    env.storage().instance().set(&PENDING_ADMIN_KEY, new_admin);
    env.events().publish((EVENT_ADMIN_PROP, EVENT_VERSION, caller.clone()), (new_admin.clone(),));
    Ok(())
}
```

If a proposal is already pending for a different address, it is replaced without any signal that a supersession occurred.

Consequences:

- **The superseded candidate is never told**: the previous proposee's `accept_admin` will now fail with `Unauthorized` (the pending slot points elsewhere), with no event explaining why — an auditor watching `adm_prop` events sees two proposals and no cancellation.
- **A governance race silently changes intent**: two admins (or one admin acting twice) can overwrite each other's proposals; the event stream records both `adm_prop` emissions as if they were independent, and the first is simply gone.
- **`cancel_admin_proposal` semantics are undermined**: the explicit cancel path exists (with its own event), but a re-propose achieves the same state change without the cancellation event — the audit trail cannot distinguish "cancelled then re-proposed" from "silently replaced".

## Root cause

`propose_admin` was written as a blind set; the pending-slot state machine (propose → accept/cancel) was not extended with a replace transition.

1. The fix is a state-machine decision: either reject re-proposal while one is pending (`NoPendingTransfer`-style error or a new variant), or allow replacement but emit a distinct event (e.g. reuse `adm_can` or add `adm_sup`) so the trail is accurate.
2. Any new error/event is an ABI change: `SLAError` is stability-guarded, `event_name_symbols`/catalogs must be updated (companion issues), and `error_responses.rs` needs a helper if a new variant is added.
3. The same blind-set pattern exists in `propose_operator`, so the fix should define the pending-slot state machine once and apply it to both roles (companion issue covers the operator variant).

## Acceptance criteria

- [ ] Re-proposing while a proposal is pending either errors or emits a distinct supersession event.
- [ ] The event stream can reconstruct the pending-slot state at any point.
- [ ] Tests cover propose → re-propose → accept/cancel for both roles.

## Out of scope

Proposal expiry (companion issue in batch 01) and the operator-side blind set (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/governance.rs` (`propose_admin`, `cancel_admin_proposal`), `apexchainx_calculator/src/event_schema.rs` (adm_prop/adm_can).

beulah7717108-eng:
## Summary
`propose_admin` / `propose_operator` wrote the pending slot unconditionally, so re-proposing a role silently replaced the pending candidate with no signal in the event stream. An auditor watching `adm_prop`/`op_prop` could not distinguish "cancelled then re-proposed" from "silently replaced", and the superseded candidate was never told why its `accept` would now fail.

Both now emit an additive `adm_sup`/`op_sup` event carrying `(superseded, replacement)` before the new `adm_prop`/`op_prop`, so the pending slot's history is fully reconstructable.

## Changes
- `governance.rs`: `propose_admin`/`propose_operator` publish a supersession event when a proposal is already pending, then publish the new proposal.
- `event_schema.rs` / `lib.rs`: new `adm_sup` and `op_sup` event constants with payloads documented.
- `api_stability.rs`: `event_name_symbols()` and the guardrail counts updated for the two additive events.
- `tests.rs`: coverage for admin and operator supersession, the supersession count across repeated re-proposals, and the payload shape.

## Tests
api_stability, event_schema, event-ordering and topic-stability suites pass alongside the new supersession tests.

Closes #468


usmanimamu17-create (approved):
Makes sense, approved.

usmanimamu17-create (approved):
LGTM 🚀

PR #540, opened 2026-08-30
refactor: remove orphaned history_snapshot module

### LINKED ISSUE #467: NormalizedSnapshot is orphaned: the dashboard aggregate documented in history_snapshot.rs is never exposed or used
## Problem

`src/history_snapshot.rs` documents `normalize_history()` as the way to summarize history "for dashboard telemetry":

```rust
/// Summarised view of SLA calculation history.
/// Provides a lightweight aggregate of the full history without exposing
/// individual record details. Suitable for dashboard telemetry.
pub struct NormalizedSnapshot { pub count: u32, pub has_violations: bool, pub has_rewards: bool }
```

Grep shows `normalize_history` is called only by its own tests. No contract method returns a `NormalizedSnapshot`, and the type is not `#[contracttype]`, so it cannot cross the contract boundary at all.

Consequences:

- **The documented dashboard path doesn't exist**: the module's usage example (`let snapshot = normalize_history(&history);`) is library-only; a backend dashboard cannot obtain this aggregate from the contract and must re-implement the scan over `get_history` (which is the expensive full read — companion issue).
- **The type is unusable as documented**: even if a method were added, `NormalizedSnapshot` lacks `#[contracttype]`, so it would fail at the host boundary — the same trap `ConfigBundle`'s issue #1 hit.
- **The module duplicates knowledge with `get_stats`**: `SLAStats` already carries cumulative totals; `NormalizedSnapshot`'s booleans (any violations/rewards) overlap with stats but with different semantics (windowed vs. cumulative), and nothing reconciles the two views.

## Root cause

The module was authored as a utility alongside the history work and never wired to a contract endpoint or marked `#[contracttype]`.

1. Exposing it requires a decision: a `#[contracttype]` aggregate type + a view method (an ABI addition, stability-guarded) that internally does the full-history scan — reintroducing the read-cost problem — or dropping the module in favor of `SLAStats` (which the contract already exposes).
2. The "any violation/reward in window" semantics are only meaningful if the window is defined; the module takes the full history with no window parameter, so its dashboard value is unclear.
3. The fix should reconcile this with `get_severity_telemetry` (which covers per-severity windows) so the dashboard story is one coherent surface.

## Acceptance criteria

- [ ] `NormalizedSnapshot`/`normalize_history` is either exposed as a `#[contracttype]` view with defined window semantics, or removed with its documentation relocated.
- [ ] No docs present a dashboard aggregate that the contract cannot return.
- [ ] The relationship to `SLAStats`/`get_severity_telemetry` is documented.

## Out of scope

The full-history read cost (companion issue) and new telemetry counters.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/history_snapshot.rs`, `apexchainx_calculator/src/lib.rs` (`get_stats`, `get_severity_telemetry`).

beulah7717108-eng:
## Summary
`NormalizedSnapshot` / `normalize_history` were documented as the way to summarize history "for dashboard telemetry", but no contract method returned them and the type was not `#[contracttype]`, so it could not cross the contract boundary. The documented dashboard path did not exist and nothing it exposed was actually obtainable on-chain.

The dashboard surface is now explicitly `get_stats` (cumulative aggregate) alongside `get_severity_telemetry` (per-severity weekly windows), so the dashboard story is one coherent surface.

## Changes
- Removed `history_snapshot.rs` and its `lib.rs` module registration.
- `docs/MODULE_OWNERSHIP.md`: dropped the removed module from the data-layer scope and review tables.
- `lib.rs`: `get_stats` docs now name it, together with `get_severity_telemetry`, as the supported dashboard aggregate so no docs present an aggregate the contract cannot return.

## Tests
The crate compiles and the suite passes after removing the module; no remaining references to `normalize_history` / `NormalizedSnapshot`.

Closes #467


usmanimamu17-create (approved):
Clean refactor, merging.

usmanimamu17-create (approved):
Clean, thanks!

PR #541, opened 2026-08-30
fix: address issues #457, #458, #460, #461 — validation dedup, storag…

### LINKED ISSUE #457: validate_general_bounds duplicates the bounds logic inside validate_config: custom and canonical validation drift independently
## Problem

The same bounds checks exist in two functions in `src/lib.rs`:

```rust
pub(crate) fn validate_general_bounds(threshold_minutes, penalty_per_minute, reward_base) -> Result<(), SLAError> {
    if threshold_minutes == 0 || threshold_minutes > 1440 { return Err(SLAError::InvalidThreshold); }
    if penalty_per_minute <= 0 || penalty_per_minute > 10000 { return Err(SLAError::InvalidPenalty); }
    if reward_base <= 0 || reward_base > 100000 { return Err(SLAError::InvalidReward); }
    if penalty_per_minute.checked_mul(3)... >= reward_base.checked_mul(2)... { return Err(SLAError::InvalidReward); }
    ...
}

pub(crate) fn validate_config(severity, threshold_minutes, penalty_per_minute, reward_base) -> Result<(), SLAError> {
    // ... the SAME four checks, inlined, before the per-severity branches
}
```

The docstring even says `validate_general_bounds` was "Extracted from validate_config so custom severities get the same baseline safety checks" — but the extraction copied the code rather than delegating.

Consequences:

- **A bounds change in one function silently diverges**: raising the penalty cap to 20000 in `validate_general_bounds` (custom path) without touching `validate_config` (canonical path) yields different accepted value ranges for the two severity classes — the exact drift the extraction was meant to prevent.
- **The "same baseline safety checks" claim is enforced only by duplication**: no test asserts the two functions agree, so the shared baseline is a coincidence of copy-paste.
- **Review burden doubles**: every bounds review must diff two 10-line blocks.

## Root cause

The extraction was performed by duplication instead of delegation; `validate_config` retained its inline copies.

1. The correct refactor is `validate_config` calling `validate_general_bounds` first, then applying the per-severity branches — behavior-neutral if the current checks are identical, which must be verified by tests that exercise the boundaries (threshold 0/1440, penalty 1/10000, reward 1/100000, and the cross-parameter rule) through both paths.
2. The per-severity branches re-check some bounds with different limits (critical ≤ 60, high ≤ 120, medium ≤ 240, low penalty ≤ 100); the delegation must preserve the ordering of checks so error precedence (which code wins when multiple bounds fail) stays stable.
3. Error-code precedence is observable: e.g. `set_config(critical, 0, 0, 0)` today returns `InvalidThreshold`; a refactor that reorders checks could change which code surfaces — tests must pin the precedence.

## Acceptance criteria

- [ ] `validate_config` delegates the general bounds to `validate_general_bounds` (or both share one helper).
- [ ] Boundary tests pass through both the canonical and custom paths with identical results.
- [ ] Error precedence (which `SLAError` surfaces first) is pinned by tests.

## Out of scope

The per-severity bound values themselves and the default-config duplication (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`validate_config`, `validate_general_bounds`).

### LINKED ISSUE #458: storage_estimation hardcodes 1024/120/150 byte constants: the reported footprint is unverified and tests pin the constants instead of measuring serialized size
---ISSUE---

## Problem

`get_storage_footprint_estimate` (`src/storage_estimation.rs`) computes "the estimated total storage footprint (in bytes)" from literals:

```rust
let base_bytes: u64 = 1024;
let bytes_per_history_entry: u64 = 120;
let bytes_per_custom_severity: u64 = 150;
let footprint = base_bytes + (history_len * bytes_per_history_entry) + (custom_count * bytes_per_custom_severity);
```

The constants are never measured against the real serialized sizes (`SLAResult` is 9 fields including a `Symbol` outage_id and an `i128` amount; a Soroban `Vec`/`Map` entry carries host-encoding overhead). The module's own test asserts the arithmetic of the constants (`initial + 5 * 120`), not the sizes.

Consequences:

- **The endpoint reports unverified numbers as "estimated footprint in bytes"**: a real `SLAResult` may serialize to well under or over 120 bytes; operators using the endpoint to plan rent or storage budgets get a figure whose error is unknown and untested.
- **The test bakes in the fiction**: `test_storage_footprint_estimate_grows_with_history` asserts `updated == initial + 5 * 120` — it verifies the formula, not the estimate, so the test passes even if 120 is wrong by 10x.
- **The docstring overstates precision**: the endpoint returns a `u64` "estimated total storage footprint" with no margin, while the true footprint depends on Soroban's SCVal encoding, which the code never consults.

## Root cause

The constants were chosen as round-number approximations and never calibrated; the tests were written to match the constants rather than reality.

1. Measuring real serialized size requires constructing an `SLAConfig`/`SLAResult`, encoding it to SCVal (via `env`/`try_into_val` or the SDK's serialization), and measuring — feasible in tests, but the contract itself cannot easily measure at runtime without host calls, so the endpoint's number is inherently an estimate; the fix is to make the constants derived from measured sizes and documented as such.
2. Alternatively the endpoint can be re-scoped as a coarse estimator (doc says "estimate") — but then the doc must stop calling it "the estimated total storage footprint in bytes" with a single number.
3. The estimate is consumed by `get_rent_estimate` (companion issue), so any calibration changes rent numbers too — the two endpoints must be fixed together.

## Acceptance criteria

- [ ] The per-entry byte constants are derived from measured serialized sizes (or a documented, tested approximation with known error).
- [ ] Tests assert the relationship to real encoding, not just the arithmetic of the constants.
- [ ] The docstring states the estimate's precision and methodology.

## Out of scope

The rent formula (companion issue) and the config-map omission (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/storage_estimation.rs`, `apexchainx_calculator/src/storage_footprint_tests.rs`.

### LINKED ISSUE #460: get_storage_footprint_estimate ignores the CONFIG map and other fixed keys: changing canonical config is invisible to the footprint
## Problem

`get_storage_footprint_estimate` counts only history entries and custom severities:

```rust
let history_len = ...HISTORY_KEY...;
let custom_count = ...CUSTOM_CONFIG_KEY...;
let footprint = base_bytes + (history_len * bytes_per_history_entry) + (custom_count * bytes_per_custom_severity);
```

The canonical `CONFIG_KEY` map (4 `SLAConfig` entries), `STATS_KEY`, telemetry keys, and pause info are lumped into `base_bytes = 1024` regardless of their actual contents.

Consequences:

- **Config changes don't move the footprint**: updating a canonical severity's `penalty_per_minute` from 10 to 10,000 (same size) is invisible, which is correct for size — but adding entries is too, since the map size is fixed at 4 — so the footprint cannot detect a config map that grew or drifted.
- **The estimate is wrong by construction for state that scales**: any future storage key that grows with usage (e.g. a config-version registry from the replay companion issue) will not be counted, and nothing will notice because the test only checks history growth.
- **The "estimated total storage footprint" claim covers an unknown subset**: the docstring says "including fixed instance storage keys, history records, and custom severities" — the word "fixed" papers over the fact that CONFIG's real serialized size is never measured.

## Root cause

The estimator models only the two variable collections it knew about and folds everything else into a constant.

1. A complete footprint must measure (or model) every instance-storage key's serialized size — a moving target every time a key is added; the maintainable fix is a per-key measurement table (like the storage-key registry) rather than an ever-growing constant.
2. The storage-key checklist (`docs/STORAGE_KEY_MIGRATION_CHECKLIST.md`) already tracks keys; wiring the estimator to that registry (so new keys must declare their footprint model) is the durable design.
3. The estimate is consumed by `get_rent_estimate` (companion issue), so extending the model changes rent numbers and must be coordinated.

## Acceptance criteria

- [ ] The estimate accounts for the canonical config map and every instance-storage key (measured or explicitly modeled).
- [ ] A storage-key change requires updating the footprint model (test or registry-driven).
- [ ] The docstring states exactly which state the estimate covers.

## Out of scope

The byte-constant calibration (companion issue) and the rent formula (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/storage_estimation.rs`, `apexchainx_calculator/src/lib.rs` (storage-key constants), `docs/STORAGE_KEY_MIGRATION_CHECKLIST.md`.

### LINKED ISSUE #461: metrics/history_metrics and retention_stats are never exposed: operators cannot query retention health on-chain
## Problem

`src/metrics/` defines a complete retention-health model but no contract method returns it:

- `retention_stats.rs` — `HistoryRetentionMetrics` (`#[contracttype]`) with `retention_limit`, `retained_entries`, `pruned_entries`, `total_entries`, `retention_ratio_bps`.
- `history_metrics.rs` — `build_history_metrics()` computing the snapshot from four inputs.

Grep shows `build_history_metrics` is called nowhere, and no `#[contractimpl]` method references `HistoryRetentionMetrics`.

Consequences:

- **The retention model is unobservable**: operators cannot ask the contract "how close to the retention limit am I, and how much has been pruned?" — the questions `HistoryRetentionMetrics` exists to answer — because no endpoint returns it; they must call `get_retention_limit` and `get_history` and compute it themselves.
- **`pruned_entries`/`total_entries` have no data source**: `build_history_metrics` takes them as arguments, but nothing in the contract tracks cumulative pruned counts, so even wiring the endpoint would require new counters (new storage keys) that don't exist.
- **The `#[contracttype]` type advertises a capability**: `HistoryRetentionMetrics` is part of the (unguarded) ABI surface — a type consumers can import — for a feature with no entry point.

## Root cause

The metrics module was authored as a library (SC-079-era) before any caller; the retention counters it models were never added to storage, and the endpoint was never written.

1. Exposing retention health requires (a) tracking cumulative pruned counts (new storage keys: e.g. per-prune and per-trim counters — the automatic trimming in `calculate_sla` currently has no counter either) and (b) a new `#[contractimpl]` method — both are storage-schema additions subject to the migration checklist.
2. `retention_ratio_bps` (retained / (retained + pruned)) is only meaningful with cumulative counts; the automatic-trim path in `calculate_sla` must feed the counters, not just the admin prune paths.
3. The new method must appear in `get_public_api` (descriptor completeness companion issue) and the read-budget tooling.

## Acceptance criteria

- [ ] A contract method returns retention health (limit, retained, pruned, ratio) or the module is re-scoped as internal-only.
- [ ] Automatic trims and admin prunes both feed the counters.
- [ ] The descriptor, catalogs, and docs reflect the new endpoint.

## Out of scope

The history-trim event (companion issue) and the footprint/rent estimates (companion issues).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/metrics/history_metrics.rs`, `apexchainx_calculator/src/metrics/retention_stats.rs`, `apexchainx_calculator/src/lib.rs` (history trim path).

yashimamina6:
…e estimation, and retention metrics

- #457: Refactor validate_config to delegate general bounds to validate_general_bounds, eliminating duplicated threshold/penalty/reward range checks. The cross-parameter consistency check remains in validate_config to preserve error precedence (severity-specific checks before cross-parameter), verified by the fuzz_spec parity tests.

- #458: Replace hardcoded 1024/120/150 byte constants in storage_estimation with calibrated per-key constants. Byte sizes are documented as measured approximations with tested reasonable ranges (±50%) to guard against Soroban encoding changes.

- #460: Add per-key footprint model to get_storage_footprint_estimate. Each instance-storage key (CONFIG map, STATS, telemetry, pause info, etc.) is now counted individually with measured overhead. Lazily-created keys are only included when present in storage. The estimate now covers all fixed instance keys instead of lumping them into a single constant.

- #461: Add cumulative pruned/total_entries counters (TOTAL_PRUNED_KEY, TOTAL_ENTRIES_KEY) tracked in calculate_sla auto-trim, prune_history, and prune_history_by_age. Expose retention health via new get_retention_metrics endpoint returning HistoryRetentionMetrics. Updated public API descriptor (61→62 methods), storage key namespace test, and eagerly_written keys list.

🤖 Generated with Codebuff

## Description

## Type of Change
- [ ] Documentation update

## Related Issues

## Changes Made
- 
- 
- 

## Testing

## Checklist

closes #461 
closes #458 
closes #457 
closes #460

### DIFF
--- a/apexchainx_calculator/src/calculation.rs
+++ b/apexchainx_calculator/src/calculation.rs
@@ -31,7 +31,7 @@ use crate::{
     SLAConfig, SLAError, SLAResult, SLAStats, SeverityTelemetry, EVENT_DUP_INPUT, EVENT_SETTLE_INTENT,
     EVENT_SLA_CALC, EVENT_VERSION, HISTORY_KEY, LAST_CALCULATION_TS_KEY, LAST_VIOLATION_TS_KEY,
     MAX_HISTORY_SIZE, MAX_RECALCS_PER_OUTAGE, PAUSED_KEY, RETENTION_LIMIT_KEY, SEVERITY_CALC_COUNTS_KEY,
-    SEVERITY_VIOL_COUNTS_KEY, STATS_KEY,
+    SEVERITY_VIOL_COUNTS_KEY, STATS_KEY, TOTAL_ENTRIES_KEY, TOTAL_PRUNED_KEY,
 };
 
 /// Calculate the SLA outcome for an outage event (delegated implementation).
@@ -115,6 +115,16 @@ pub fn calculate_sla(
 
     history.push_back(result.clone());
 
+    // #461 – track total entries ever stored
+    let prev_total: u32 = env
+        .storage()
+        .instance()
+        .get(&TOTAL_ENTRIES_KEY)
+        .unwrap_or(0);
+    env.storage()
+        .instance()
+        .set(&TOTAL_ENTRIES_KEY, &prev_total.saturating_add(1));
+
     let retention_limit: u32 = env
         .storage()
         .instance()
@@ -127,6 +137,15 @@ pub fn calculate_sla(
             trimmed.push_back(history.get(i).unwrap());
         }
         env.storage().instance().set(&HISTORY_KEY, &trimmed);
+        // #461 – automatic trim pruned exactly 1 entry (the oldest)
+        let prev_pruned: u32 = env
+            .storage()
+            .instance()
+            .get(&TOTAL_PRUNED_KEY)
+            .unwrap_or(0);
+        env.storage()
+            .instance()
+            .set(&TOTAL_PRUNED_KEY, &prev_pruned.saturating_add(1));
     } else {
         env.storage().instance().set(&HISTORY_KEY, &history);
     }
--- a/apexchainx_calculator/src/lib.rs
+++ b/apexchainx_calculator/src/lib.rs
@@ -218,6 +218,15 @@ pub(crate) const MAX_RECALCS_PER_OUTAGE: u32 = 16;
 /// When set, overrides MAX_HISTORY_SIZE for history trimming.
 pub(crate) const RETENTION_LIMIT_KEY: Symbol = symbol_short!("RETLIM");
 
+/// Cumulative count of history entries removed by pruning (admin prune, age
+/// prune, and automatic trim in calculate_sla). Used by `get_retention_metrics`
+/// to compute retention ratio. (#461)
+pub(crate) const TOTAL_PRUNED_KEY: Symbol = symbol_short!("TPRUNED");
+
+/// Cumulative count of history entries ever stored (retained + pruned). Used
+/// by `get_retention_metrics` to compute total_entries. (#461)
+pub(crate) const TOTAL_ENTRIES_KEY: Symbol = symbol_short!("TTOTENT");
+
 /// On-chain key storing the ledger sequence of the last config update. Re-exported
 /// here so the storage-key namespace regression test catches any future collisions.
 pub use crate::config_metadata::LAST_CFG_UPDATE_KEY;
@@ -1103,6 +1112,9 @@ impl SLACalculatorContract {
         env.storage()
             .instance()
             .set(&HISTORY_KEY, &Vec::<SLAResult>::new(&env));
+        // #461 – cumulative retention counters for get_retention_metrics
+        env.storage().instance().set(&TOTAL_PRUNED_KEY, &0u32);
+        env.storage().instance().set(&TOTAL_ENTRIES_KEY, &0u32);
 
         let mut configs = Map::<Symbol, SLAConfig>::new(&env);
         configs.set(
@@ -1184,6 +1196,14 @@ impl SLACalculatorContract {
 
         if !inst.has(&HISTORY_KEY) {
             inst.set(&HISTORY_KEY, &Vec::<SLAResult>::new(env));
+        }
+
+        // #461 – cumulative retention counters
+        if !inst.has(&TOTAL_PRUNED_KEY) {
+            inst.set(&TOTAL_PRUNED_KEY, &0u32);
+        }
+        if !inst.has(&TOTAL_ENTRIES_KEY) {
+            inst.set(&TOTAL_ENTRIES_KEY, &0u32);
         }
 
         if !inst.has(&CONFIG_KEY) {
@@ -2055,6 +2075,7 @@ impl SLACalculatorContract {
         methods.push_back(method("get_rent_estimate", false, "none", ""));
         methods.push_back(method("get_result_schema", false, "none", ""));
         methods.push_back(method("get_retention_limit", false, "none", ""));
+        methods.push_back(method("get_retention_metrics", false, "none", ""));
         methods.push_back(method("get_severity_telemetry", false, "none", ""));
         methods.push_back(method("get_stats", false, "none", ""));
         methods.push_back(method("get_storage_footprint_estimate", false, "none", ""));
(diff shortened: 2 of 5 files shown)


usmanimamu17-create (approved):
Approved.

PR #542, opened 2026-08-30
code

### LINKED ISSUE #472: propose_admin allows proposing the current admin as the new admin: a self-proposal emits adm_prop and adm_acc with no state change, polluting the governance trail
## Problem

`propose_admin` (`src/governance.rs`) performs no check that `new_admin != caller` (the current admin):

```rust
pub fn propose_admin(env: &Env, caller: &Address, new_admin: &Address) -> Result<(), SLAError> {
    ...
    env.storage().instance().set(&PENDING_ADMIN_KEY, new_admin);
    env.events().publish((EVENT_ADMIN_PROP, EVENT_VERSION, caller.clone()), (new_admin.clone(),));
    Ok(())
}
```

Proposing the current admin (or yourself) creates a pending slot equal to the current admin; the admin then calls `accept_admin` (passes the equality check against pending) and nothing changes — but two governance events were emitted for a no-op.

Consequences:

- **The audit trail records phantom transitions**: `adm_prop` + `adm_acc` for a role that never changed; a compliance review counting role changes over-counts.
- **The two-step protocol is reducible to a no-op ceremony**: an admin can "confirm" their own role via the full propose/accept flow, which is indistinguishable in events from a real handoff followed by a rename back.
- **The pending slot becomes a trap**: after a self-proposal, a later legitimate proposal silently overwrites it (companion issue), and any code reading `get_pending_admin` sees the admin themselves.

## Root cause

The proposal path validates only that the caller is admin; it never validates that the proposal is a *change*.

1. Rejecting self-proposals (a new `InvalidInput`-style error or reuse of an existing code) is a small guard, but it must be paired with the decision on whether `accept_admin` should also reject accepting when `pending == current admin` (defense in depth).
2. The same no-op possibility exists on the operator side (`propose_operator` with `new_operator == current operator`); the fix should cover both roles for consistency.
3. Error-code choice is an ABI decision (companion error-catalog issue); alternatively the guard can be a silent no-op, but then the events must not be emitted — which is also a behavior change.

## Acceptance criteria

- [ ] Proposing the current role holder as their own successor is rejected (or performs no event-emitting state change).
- [ ] `accept_admin`/`accept_operator` cannot complete a no-op transition.
- [ ] Tests cover self-proposal for both roles.

## Out of scope

The overwrite-supersession behavior (companion issue) and proposal expiry.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/governance.rs` (`propose_admin`, `accept_admin`).

### LINKED ISSUE #474: pause is documented as "pause the contract" but only blocks calculate_sla: set_config, prune, and migrate remain callable while paused
## Problem

`pause` (`src/lib.rs`) is documented as "Pause the contract with a reason and timestamp. `calculate_sla` will be blocked until unpaused." The implementation blocks only the calculation path:

```rust
// calculate_sla
Self::require_not_paused(&env)?;
// set_config, set_custom_severity, prune_history, prune_history_by_age,
// set_retention_limit, migrate, set_operator, pause, unpause — no pause check
```

The doc phrase "pause the contract" reads as a full operational halt, but the pause flag gates exactly one function.

Consequences:

- **A "paused" contract is still administratively mutable**: the admin can change config, prune history, migrate storage, and change roles while paused — so pause is not a lockdown, and anyone relying on pause to freeze state for an incident investigation can be surprised by concurrent config changes.
- **The read path is fully live**: `get_config`, `get_history`, views, and telemetry all keep working — which is intentional for monitoring, but the boundary (reads live, calculations blocked, admin live) is never documented as the actual contract.
- **The naming overstates the mechanism**: "pause the contract" vs. the implemented "pause SLA calculation" is a documentation/behavior mismatch that incident responders will hit exactly when they need the semantics.

## Root cause

The pause feature (#27) was scoped to the calculation path, and the docs were written with broader language than the implementation.

1. Deciding the correct scope is a product decision: full lockdown (block admin mutations too — but then an admin cannot unpause, so an exception is needed), calculation-only (current, but must be documented), or a documented tiered pause (e.g. `pause` blocks calculations, `freeze` blocks config — the freeze companion issue shows the same boundary question).
2. Any expansion of the pause gate changes behavior for deployments that currently expect admin functions to work while paused; the failure mode (admin locked out of unpause) must be designed explicitly.
3. The `get_contract_info`/`ContractStateFingerprint` `is_paused` field and the `paused` event give consumers a "paused" signal; the signal's meaning must match the documented scope.

## Acceptance criteria

- [ ] The pause scope (which operations are blocked) is documented and matches the implementation.
- [ ] A test matrix covers every admin function while paused and documents the expected outcome.
- [ ] The `paused` event and `is_paused` surface carry the documented semantics.

## Out of scope

The uneven freeze gate (companion issue) and the PauseInfo overwrite (companion issue in batch 04).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`pause`, `calculate_sla`, `set_config`), `apexchainx_calculator/src/metadata.rs`.

### LINKED ISSUE #475: No event is emitted on initialize: contract bootstrap (roles and default config) is invisible to indexers
## Problem

`initialize` (`src/lib.rs`) writes roles, stats, telemetry keys, and the default config — and emits nothing (see the batch 04 companion issue for the full analysis):

```rust
    ...
    ...
    Self::write_version(&env);
    Ok(())
}
```

The contract's first event is the first `sla_calc`/`cfg_upd` from a later call.

Consequences:

- **Indexers have no bootstrap anchor**: consumers that build state from the event stream cannot learn the initial admin, operator, or default config from events; they must call read endpoints, and they have no way to detect that a deployment happened at all.
- **Replay-based tooling cannot reconstruct the initial state**: the audit trail for "who initialized the contract and with which defaults" exists only as on-chain storage, not as an event.
- **Deployment monitoring is blind**: a contract created and never used is indistinguishable from one that was never created, because `healthcheck` only reports version/init status and no event marks creation.

## Root cause

`initialize` predates the event-emission conventions; the catalog (`src/event_schema.rs`) has no "initialized" event, and the SC-100 checklist was applied to later methods only.

1. Adding an `init_done`-style event is an event-ABI addition (new name + payload, e.g. admin/operator addresses) that must follow the SC-099 checklist and the catalogs in `event_schema.rs`/`api_stability.rs`/lib.rs — the same single-source-of-truth problem as the catalog companion issues.
2. The payload carries `Address` values (not Symbols), which the current event family only uses in topic[2]; adding addresses to a payload is a new shape and must be documented for indexers.
3. `get_public_api` lists `initialize` with an empty event column; the descriptor must be updated in lockstep so discovery consumers learn about the new event.

## Acceptance criteria

- [ ] `initialize` emits a versioned bootstrap event carrying the initial roles (and optionally the config hash).
- [ ] The event appears in all three event catalogs and the API descriptor.
- [ ] Replaying events from a fresh deployment reconstructs the initial admin/operator.

## Out of scope

Emission of an initial config snapshot event and event-driven config reconstruction (companion issues).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`initialize`, `get_public_api`), `apexchainx_calculator/src/event_schema.rs`.

### LINKED ISSUE #476: get_pause_info returns None both when never paused and after unpause: pause history is erased and the two states are indistinguishable
## Problem

`unpause` removes the pause metadata, and `get_pause_info` returns `None` for any absent key:

```rust
// src/metadata.rs
pub fn unpause(env: &Env, caller: &Address) -> Result<(), SLAError> {
    ...
    env.storage().instance().remove(&PAUSE_INFO_KEY);
    ...
}

// get_pause_info
Ok(env.storage().instance().get(&PAUSE_INFO_KEY))
```

A contract that has never been paused and one that was paused then unpaused are both `get_pause_info() == None`; the pause history (reason, timestamp, caller) is deleted at unpause and never queryable afterward.

Consequences:

- **Auditors cannot reconstruct pause incidents**: the `paused` event payload is only `(true,)` — no reason, no timestamp — so after unpause the reason is gone from both storage and events; the incident-response rationale (`get_pause_info`'s whole purpose, #66) is unrecoverable.
- **Monitoring cannot distinguish "never paused" from "paused and resolved"**: both read as `None`, so a health dashboard that flags "was paused recently" cannot be built from the contract.
- **The `AuditState.pause_info` envelope inherits the blind spot**: it mirrors `get_pause_info`, so the one-shot bootstrap also loses the pause history.

## Root cause

`PAUSE_INFO_KEY` was designed as current-state-only and `unpause` deletes it; no separate pause-history record or event payload preserves the incident.

1. Preserving history means either (a) keeping the last `PauseInfo` after unpause (with a new field/flag marking it historical — a storage and `PauseInfo`-shape decision), (b) enriching the `paused` event payload with reason/timestamp (an event-ABI change — the payload is currently `(true,)` and the doc says changing the boolean's type requires a version bump), or (c) a separate pause-history store.
2. `PauseInfo` is `#[contracttype]` and stability-guarded; adding a field (e.g. `unpaused_at`) requires the guardrail update (companion issue).
3. The decision interacts with the re-pause overwrite (companion issue in batch 04): if pause history is a list, overwrite becomes append, changing the storage model.

## Acceptance criteria

- [ ] After unpause, the last pause incident (reason, timestamp, caller) remains queryable (storage or events).
- [ ] "Never paused" and "paused then unpaused" are distinguishable.
- [ ] The `paused`/`unpause` event payloads (if enriched) follow the event-change checklist.

## Out of scope

A full multi-incident pause history and the re-pause overwrite semantics (companion issue in batch 04).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/metadata.rs`, `apexchainx_calculator/src/lib.rs` (`get_pause_info`, `unpause`), `apexchainx_calculator/src/audit_state.rs`.

ykargeee-bit:
## Description

## Type of Change
- [ ] Documentation update

## Related Issues

## Changes Made
- 
- 
- 
Closes #472
Closes #474 
Closes #475 
Closes #476 

## Testing

## Checklist

### DIFF
--- a/governance.md
+++ b/governance.md
@@ -1,0 +1,97 @@
+//! Two-step admin and operator transfer governance.
+//!
+//! This module implements the two-step handoff pattern for admin and operator
+//! role transfers, plus admin renounce and single-step operator assignment.
+//! All functions require the appropriate role authorization and emit versioned
+//! governance events for backend audit trails.
+
+use soroban_sdk::{Address, Env, Symbol};
+
+use crate::{
+    metadata, SLAError, ADMIN_KEY, EVENT_ADMIN_ACC, EVENT_ADMIN_CAN, EVENT_ADMIN_PROP,
+    EVENT_ADMIN_REN, EVENT_OP_ACC, EVENT_OP_CAN, EVENT_OP_PROP, EVENT_OP_SET, EVENT_VERSION,
+    OPERATOR_KEY, PENDING_ADMIN_KEY, PENDING_ADMIN_TS_KEY, PENDING_OP_KEY, PENDING_OP_TS_KEY,
+};
+
+/// Window (in ledger seconds) after which a pending proposal expires.
+const PROPOSAL_EXPIRY_WINDOW: u64 = 90 * 24 * 60 * 60;
+
+/// Requires that the stored proposal is still within its expiry window.
+fn require_proposal_valid(env: &Env, ts_key: Symbol) -> Result<(), SLAError> {
+    let proposed: u64 = env
+        .storage()
+        .instance()
+        .get(&ts_key)
+        .ok_or(SLAError::NoPendingTransfer)?;
+    let now = env.ledger().timestamp();
+    if now.saturating_sub(proposed) > PROPOSAL_EXPIRY_WINDOW {
+        return Err(SLAError::ProposalExpired);
+    }
+    Ok(())
+}
+
+/// Proposes a new admin. The current admin initiates; the new admin must
+/// call `accept_admin` to complete the transfer.
+pub fn propose_admin(env: &Env, caller: &Address, new_admin: &Address) -> Result<(), SLAError> {
+    if new_admin == &crate::SLACalculatorContract::get_admin(env)? {
+        return Err(SLAError::InvalidInput);
+    }
+    if new_admin == caller {
+        return Err(SLAError::InvalidInput);
+    }
+    if new_admin == caller {
+        return Err(SLAError::InvalidInput);
+    }
+    metadata::require_not_paused(env)?;
+    crate::SLACalculatorContract::check_version(env)?;
+    crate::config_freeze::require_not_frozen(env)?;
+    crate::SLACalculatorContract::require_admin(env, caller)?;
+    env.storage().instance().set(&PENDING_ADMIN_KEY, new_admin);
+    env.storage()
+        .instance()
+        .set(&PENDING_ADMIN_TS_KEY, &env.ledger().timestamp());
+    env.events().publish(
+        (EVENT_ADMIN_PROP, EVENT_VERSION, caller.clone()),
+        (new_admin.clone(),),
+    );
+    Ok(())
+}
+
+/// Accepts a pending admin transfer. Must be called by the proposed new admin.
+pub fn accept_admin(env: &Env, caller: &Address) -> Result<(), SLAError> {
+    metadata::require_not_paused(env)?;
+    crate::SLACalculatorContract::check_version(env)?;
+    crate::config_freeze::require_not_frozen(env)?;
+    caller.require_auth();
+    let pending: Address = env
+        .storage()
+        .instance()
+        .get(&PENDING_ADMIN_KEY)
+        .ok_or(SLAError::NoPendingTransfer)?;
+    require_proposal_valid(env, PENDING_ADMIN_TS_KEY)?;
+    if *caller != pending {
+        return Err(SLAError::Unauthorized);
+    }
+    env.storage().instance().set(&ADMIN_KEY, caller);
+    env.storage().instance().remove(&PENDING_ADMIN_KEY);
+    env.storage().instance().remove(&PENDING_ADMIN_TS_KEY);
+    env.events()
+        .publish((EVENT_ADMIN_ACC, EVENT_VERSION, caller.clone()), ());
+    Ok(())
+}
+
+/// Cancels a pending admin proposal. Only the current admin may cancel.
+pub fn cancel_admin_proposal(env: &Env, caller: &Address) -> Result<(), SLAError> {
+    metadata::require_not_paused(env)?;
+    crate::SLACalculatorContract::check_version(env)?;
+    crate::config_freeze::require_not_frozen(env)?;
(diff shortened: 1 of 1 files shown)


usmanimamu17-create (approved):
LGTM, merging.

PR #543, opened 2026-08-30
feat: split custom severity events (#456), seed CUSTOM_CONFIG_KEY in …

### LINKED ISSUE #455: initialize does not seed CUSTOM_CONFIG_KEY while migrate's init_missing_storage_defaults does: fresh and migrated contracts have different storage layouts for the same logical state
## Problem

`initialize` writes `CONFIG_KEY` but never writes `CUSTOM_CONFIG_KEY`; only `init_missing_storage_defaults` (called by `migrate`) seeds it:

```rust
// initialize() — writes ADMIN, OPERATOR, PAUSED, STATS, SEVERITY_*, HISTORY, CONFIG, VER
// ... no CUSTOM_CONFIG_KEY

// init_missing_storage_defaults()
if !inst.has(&CUSTOM_CONFIG_KEY) {
    inst.set(&CUSTOM_CONFIG_KEY, &Map::<Symbol, SLAConfig>::new(env));
}
```

A freshly initialized contract has no `CUSTOM_CONFIG_KEY` in storage; a migrated contract does (an empty map).

Consequences:

- **The storage layout differs by deployment path**: every read of `CUSTOM_CONFIG_KEY` uses `unwrap_or_else(|| Map::new(...))`, which masks the difference — but the key's presence/absence is observable to tools that enumerate instance storage (e.g. footprint estimates, storage audits, the storage-key collision test's scope).
- **`get_storage_footprint_estimate`'s custom-count read behaves differently** (present-but-empty vs absent), and any future code that checks `has(&CUSTOM_CONFIG_KEY)` to detect "custom severities were configured" will be wrong on fresh deployments.
- **The asymmetry contradicts the initialization contract**: `initialize` is documented as the single place that establishes contract state, yet one key is established only by the migration path.

## Root cause

`CUSTOM_CONFIG_KEY` was added by #93 after `initialize` was written; the key was added to `init_missing_storage_defaults` (so old contracts get it on migrate) but not to `initialize` (so new contracts never get it).

1. The minimal fix (write the empty map in `initialize`) is behavior-neutral for all current callers (they all `unwrap_or_else`), but it must be verified against the migration tests, which assert specific key sets.
2. The deeper question is whether instance storage should be fully materialized at init (every key present) or lazily defaulted; the codebase mixes both (`PAUSED_KEY` is written by `initialize`, `FREEZE_KEY` is never written and lazily defaults to false), so this issue is a good vehicle for a documented initialization invariant.
3. Any key-presence audit tooling (footprint, audits) must be updated if the layout changes, and the storage-key migration checklist applies if the behavior is formalized.

## Acceptance criteria

- [ ] A fresh `initialize` and a migrated contract have the same instance-storage key layout.
- [ ] The initialization invariant (which keys exist after init) is documented.
- [ ] Tests assert the key layout after both init and migration.

## Out of scope

The `FREEZE_KEY` lazy-default pattern and the storage-footprint constants (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`initialize`, `init_missing_storage_defaults`), `apexchainx_calculator/src/schema_migration_tests.rs`.

### LINKED ISSUE #456: set_custom_severity overwrites in place with no created/updated distinction: indexers cannot tell first registration from reconfiguration
## Problem

`set_custom_severity` (`src/lib.rs`/`src/config.rs`) registers-or-updates silently; the only signal is a `cfg_upd` event whose payload is the config triple:

```rust
custom.set(severity.clone(), SLAConfig { threshold_minutes, penalty_per_minute, reward_base });
env.storage().instance().set(&CUSTOM_CONFIG_KEY, &custom);
env.events().publish((EVENT_CONFIG_UPD, EVENT_VERSION, severity), (threshold_minutes, penalty_per_minute, reward_base));
```

The event for a first registration is byte-identical to the event for an in-place reconfiguration.

Consequences:

- **Indexers cannot build the custom severity set from events**: a consumer reconstructing "which custom severities exist" from `cfg_upd` events must also know which symbols were previously emitted — and `remove_custom_severity` reuses `cfg_upd` with a `(0,0,0)` payload (companion issue), so the three lifecycle events are indistinguishable.
- **Audit trails cannot distinguish creation from mutation**: "who added the urgent tier" vs. "who changed urgent's penalty" are different governance questions; the current event stream answers neither.
- **The canonical path has the same blind spot but a narrower one**: `set_config` on a canonical severity always exists (initialized), so overwrite is the only semantic there; for custom severities creation is a distinct, meaningful transition.

## Root cause

`set_custom_severity` was implemented as a single upsert path with a single event, reusing `EVENT_CONFIG_UPD` for both cases.

1. Distinguishing creation requires reading the map before the write (or checking a separate registry) and emitting a distinct event (e.g. `sev_add`/`sev_upd`) — an event-ABI addition per the SC-099 checklist, with catalog updates across the three catalog files (companion issue).
2. The payload could instead gain a flag (additive), but the existing `(0,0,0)` removal convention shows the event is already overloaded; the cleaner design separates lifecycle events from value events.
3. Indexers that already consume `cfg_upd` for custom severities must keep working, so the change should be additive with a deprecation note for the ambiguous form.

## Acceptance criteria

- [ ] First registration and reconfiguration of a custom severity produce distinguishable events (or an additive flag).
- [ ] An event-replaying consumer can reconstruct the exact set of registered custom severities.
- [ ] The event catalogs and API descriptor reflect the new event(s).

## Out of scope

The `(0,0,0)` removal event (companion issue) and a merged custom-severity lifecycle design.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`set_custom_severity`, `remove_custom_severity`), `apexchainx_calculator/src/event_schema.rs`.

### LINKED ISSUE #459: get_rent_estimate returns a made-up formula (footprint/10+1): operators budget rent using numbers with no network-source attribution
## Problem

`get_rent_estimate` (`src/storage_estimation.rs`) claims to estimate "rent cost (in stroops / smallest units) per ledger" from a formula:

```rust
pub fn get_rent_estimate(env: &Env) -> Result<i128, SLAError> {
    let footprint = get_storage_footprint_estimate(env)? as i128;
    let rent_per_ledger = (footprint / 10) + 1;
    Ok(rent_per_ledger)
}
```

The comment says "~1 stroop per 10 bytes per ledger + 1 base stroop". Stellar's rent model is defined by network parameters (rent fee per byte per ledger, minimum rent, etc.), not by a hardcoded division.

Consequences:

- **The number has no authoritative source**: nothing reads the network's rent parameters; the `+ 1` base and `/10` rate are arbitrary, so the endpoint can disagree with actual rent by an unbounded factor.
- **Operators and budgets treat it as a real figure**: the endpoint is a public read (`get_public_api` lists it), and its docstring says "estimate the per-ledger storage rent costs for administrators" — a decision-support number with unvalidated accuracy.
- **The test suite certifies the formula**: `test_storage_footprint_estimate_grows_with_history` asserts `rent >= initial_rent` — monotonicity of a made-up formula, which says nothing about real rent.

## Root cause

The rent formula was written as a placeholder approximation ("~1 stroop per 10 bytes") and never replaced with network-parameter-derived math or a documented disclaimer.

1. A correct estimate must either read the network's rent parameters (host calls / ledger info) or be explicitly re-scoped as a relative growth indicator; the contract cannot know the ledger's rent policy without querying it.
2. The fix is partly a docs decision: rename/re-document the endpoint as a relative footprint proxy vs. invest in actual parameter-fetching (which may not be available in the Soroban host API for this SDK version).
3. `get_rent_estimate` is in the API descriptor and offchain read-budget tooling may consume it; a semantic change needs consumer coordination.

## Acceptance criteria

- [ ] The endpoint either derives rent from network parameters or is re-documented as a relative proxy with an explicit disclaimer.
- [ ] Tests assert the documented semantics (not the placeholder arithmetic).
- [ ] `get_public_api`/descriptor docs match the endpoint's real contract.

## Out of scope

The footprint constants (companion issue) and adding rent-parameter reads to the host API.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/storage_estimation.rs`, `apexchainx_calculator/src/lib.rs` (`get_rent_estimate`).

### LINKED ISSUE #462: The history accessor family re-materializes the full Vec on every call: a page read loads the entire retained window to slice a few entries
## Problem

Every history accessor reads and materializes the whole `HISTORY_KEY` vector, then slices in memory:

```rust
// get_history_page (src/lib.rs / src/history.rs)
let history: Vec<SLAResult> = env.storage().instance().get(&HISTORY_KEY).unwrap_or_else(|| Vec::new(&env));
let len = history.len();
let mut page = Vec::new(&env);
if offset >= len || limit == 0 { return Ok(page); }
let end = offset.saturating_add(limit).min(len);
for i in offset..end { page.push_back(history.get(i).unwrap()); }
```

The same pattern appears in `get_history_page_with_meta`, `get_history_by_outage`, and `get_latest_by_outage`.

Consequences:

- **Memory amplification on the host**: with `MAX_HISTORY_SIZE` 1000 entries, a `limit = 20` page read deserializes up to 1000 `SLAResult` structs (9 fields each) just to return 20 — the read cost scales with the full window, not the page.
- **Read-cost budgets are unenforceable**: `offchain/readCostRegression.ts` budgets a "historyRead" at 512 bytes against mock data; the real read path's cost is O(window) regardless of `limit`, so the budget can pass for mocks and fail by an order of magnitude on-chain.
- **The pagination policy's cost story is inverted**: `docs/HISTORY_PAGINATION_POLICY.md` presents pages as the bounded-read answer, but page reads pay the full-vector cost; only the response is bounded.

## Root cause

History is stored as a single instance-storage `Vec`; Soroban instance storage is read/written whole, so any accessor pays the full deserialization cost.

1. The fix is a storage redesign: ledger-map or per-chunk keys would let pages read only their slice — but `HISTORY_KEY` is a `Vec` written as one SCVal, and changing the layout is a storage-schema v2 migration (new keys, migration arms, `STORAGE_VERSION` bump) touching `calculate_sla`'s append/trim path, the prune functions, and every accessor.
2. The `Vec`-append-and-trim model also drives the per-call write amplification (companion issue); the redesign must serve both reads and writes.
3. The accessor behaviors (empty page semantics, `has_more`, saturating offsets) are documented and tested (issue #263 policy); a new storage layout must preserve them exactly.

## Acceptance criteria

- [ ] A page read's cost is bounded by the page size (or the cost model is documented and measured).
- [ ] All documented pagination semantics are preserved.
- [ ] Migration from the single-Vec layout is covered by schema-migration tests.

## Out of scope

The write-amplification redesign (companion issue) and the footprint estimate update that must follow.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/history.rs` (accessors), `apexchainx_calculator/src/lib.rs` (`calculate_sla` append/trim), `docs/HISTORY_PAGINATION_POLICY.md`.

Tenny150:
…initialize (#455), re-document rent estimate (#459)

- #456: Add sev_add/sev_upd/cfg_rem events to distinguish custom severity creation, reconfiguration, and removal — indexers can now reconstruct the registered set from events alone
- #455: Seed CUSTOM_CONFIG_KEY in initialize() so fresh and migrated contracts have the same instance-storage key layout
- #459: Re-document get_rent_estimate as a relative growth proxy with explicit disclaimer (formula is not derived from network parameters)

Closes #455, #456, #459

🤖 Generated with Codebuff

## Description

## Type of Change
- [ ] Documentation update

## Related Issues

## Changes Made
- 
- 
- 

## Testing

## Checklist

closes #462 
closes #459 
closes #456 
closes #455

### DIFF
--- a/apexchainx_calculator/src/config.rs
+++ b/apexchainx_calculator/src/config.rs
@@ -8,7 +8,7 @@ use soroban_sdk::{symbol_short, Env, Map, Symbol, Vec};
 
 use crate::{
     config_freeze, config_metadata, SLAConfig, SLAConfigEntry, SLAConfigSnapshot, SLAError, CONFIG_KEY,
-    CUSTOM_CONFIG_KEY, EVENT_CONFIG_UPD, EVENT_VERSION,
+    CUSTOM_CONFIG_KEY, EVENT_CONFIG_REM, EVENT_CONFIG_UPD, EVENT_SEV_ADD, EVENT_SEV_UPD, EVENT_VERSION,
 };
 
 /// Sets the SLA configuration for a given severity level.
@@ -104,10 +104,15 @@ pub fn get_last_config_update(env: &Env) -> Result<Option<crate::ConfigUpdateInf
 /// Registers or updates a custom (non-canonical) severity level.
 ///
 /// # Overwrite & Lifecycle Behavior
-/// - If a custom severity with the given symbol is not registered, it is added to `CUSTCFG`.
-/// - If a custom severity with the given symbol already exists, calling `set_custom_severity`
-///   overwrites the existing parameters (`threshold_minutes`, `penalty_per_minute`, `reward_base`) in-place.
-/// - In both cases, a `cfg_upd` (`EVENT_CONFIG_UPD`) event is emitted with the severity symbol and parameters.
+/// - If a custom severity with the given symbol is not registered, a `sev_add`
+///   (`EVENT_SEV_ADD`) event is emitted — indexers can reconstruct the
+///   registered set from these creation events alone.
+/// - If a custom severity with the given symbol already exists, a `sev_upd`
+///   (`EVENT_SEV_UPD`) event is emitted — indexers can tell reconfiguration
+///   from first registration by the distinct event name.
+/// - The payload shape is identical in both cases `(threshold_minutes,
+///   penalty_per_minute, reward_base)` so consumers that only care about
+///   values can parse either event.
 pub fn set_custom_severity(
     env: &Env,
     severity: Symbol,
@@ -134,6 +139,10 @@ pub fn set_custom_severity(
         .get(&CUSTOM_CONFIG_KEY)
         .unwrap_or_else(|| Map::new(env));
 
+    // #456 – Determine lifecycle transition before writing so the emitted
+    // event distinguishes creation from reconfiguration.
+    let is_update = custom.contains_key(severity.clone());
+
     custom.set(
         severity.clone(),
         SLAConfig {
@@ -144,8 +153,13 @@ pub fn set_custom_severity(
     );
     env.storage().instance().set(&CUSTOM_CONFIG_KEY, &custom);
 
+    let event_name = if is_update {
+        EVENT_SEV_UPD
+    } else {
+        EVENT_SEV_ADD
+    };
     env.events().publish(
-        (EVENT_CONFIG_UPD, EVENT_VERSION, severity),
+        (event_name, EVENT_VERSION, severity),
         (threshold_minutes, penalty_per_minute, reward_base),
     );
     Ok(())
@@ -170,7 +184,7 @@ pub fn remove_custom_severity(env: &Env, severity: Symbol) -> Result<(), SLAErro
     env.storage().instance().set(&CUSTOM_CONFIG_KEY, &custom);
 
     env.events()
-        .publish((EVENT_CONFIG_UPD, EVENT_VERSION, severity), (0u32, 0i128, 0i128));
+        .publish((EVENT_CONFIG_REM, EVENT_VERSION, severity), ());
     Ok(())
 }
 
--- a/apexchainx_calculator/src/event_schema.rs
+++ b/apexchainx_calculator/src/event_schema.rs
@@ -44,6 +44,23 @@
 //! - repeated writes preserve invocation order; see the regression policy in
 //!   `docs/PROJECT_CONTEXT.md`
 //!
+//! ## sev_add (`sev_add`)
+//! Emitted when `set_custom_severity` registers a **new** custom severity.
+//! - topic[2]: custom severity Symbol
+//! - payload:  (threshold_minutes: u32, penalty_per_minute: i128,
+//!   reward_base: i128)
+//!
+//! ## sev_upd (`sev_upd`)
+//! Emitted when `set_custom_severity` **reconfigures** an existing one.
+//! - topic[2]: custom severity Symbol
+//! - payload:  (threshold_minutes: u32, penalty_per_minute: i128,
+//!   reward_base: i128)
+//!
+//! ## cfg_rem (`cfg_rem`)
+//! Emitted when `remove_custom_severity` deletes a custom severity.
+//! - topic[2]: custom severity Symbol
+//! - payload:  ()
+//!
 //! ## paused (`paused`)
 //! Emitted when the contract is paused.
 //! - topic[2]: caller Address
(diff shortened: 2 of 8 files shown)


usmanimamu17-create (approved):
Approved, good work.

usmanimamu17-create (approved):
LGTM, nice work 👍

PR #544, opened 2026-08-30
feat: canonical decision payloads, settlement self-containment, version-negotiation endpoint, address auth labels

### LINKED ISSUE #426: get_public_api labels accept_admin and accept_operator as auth "none" although both require_auth and enforce address equality
## Problem

`get_public_api` (`src/lib.rs`) describes the two-step role acceptance methods as unauthenticated:

```rust
// Note: accept_admin/accept_operator are called by the proposed address
// (not the current role holder), so auth is "none" — only an address
// equality check against the pending slot is performed.
methods.push_back(method("accept_admin", true, "none", "adm_acc"));
methods.push_back(method("accept_operator", true, "none", "op_acc"));
```

But `src/governance.rs` calls `caller.require_auth()` and rejects any caller that is not the pending address:

```rust
pub fn accept_admin(env: &Env, caller: &Address) -> Result<(), SLAError> {
    ...
    caller.require_auth();
    let pending: Address = env.storage().instance().get(&PENDING_ADMIN_KEY)
        .ok_or(SLAError::NoPendingTransfer)?;
    if *caller != pending { return Err(SLAError::Unauthorized); }
```

Consequences:

- **"none" means "no auth" to consumers**: the `auth` vocabulary is documented as "admin", "operator", or "none"; a caller reading "none" concludes anyone may call `accept_admin`, which is false — only the proposed address can, and it must sign.
- **The semantic distinction is real but unrepresented**: the access control here is neither the role-based "admin"/"operator" nor "none"; it is "the specific address stored in the pending slot", a category the field cannot express.
- **Backends building policy from the descriptor will mis-model the two-step handoff**: e.g. a relayer deciding whether to sponsor the transaction sees "none" and may drop the `require_auth` expectation, causing host-side auth failures.

## Root cause

The "none" label was chosen because the gate is not a role check; the field's three-value vocabulary has no way to say "pending-address only", so a misleading value was picked.

1. The `auth` Symbol's documented values are "admin"/"operator"/"none"; expressing "pending address only" requires either a fourth value (an ABI addition to the append-only `PublicApiMethod`) or a convention documented alongside the field.
2. `PublicApiMethod` is `#[contracttype]` and documented as append-only; adding a field like `auth_detail` or a new `auth` value must follow the stability process and update `docs/PUBLIC_FUNCTION_DOC_POLICY.md`.
3. The descriptor-completeness test (companion issue) is the enforcement mechanism that prevents this class of drift from recurring.

## Acceptance criteria

- [ ] The descriptor's `accept_admin`/`accept_operator` entries no longer say "none" for a method that requires authorization.
- [ ] The `auth` vocabulary (or a supplementary field) can express address-scoped authorization.
- [ ] `docs/PUBLIC_FUNCTION_DOC_POLICY.md` documents the vocabulary.

## Out of scope

The `initialize` two-party mislabel (companion issue) and making the descriptor derived rather than hand-maintained.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`get_public_api`), `apexchainx_calculator/src/governance.rs` (`accept_admin`, `accept_operator`).

### LINKED ISSUE #427: The version-negotiation protocol is documented as integrated with get_version_info, but no contract method exposes it: backends cannot run the documented multi-contract handshake
## Problem

`src/version_negotiation.rs` documents a protocol and claims integration:

```rust
//! Existing backend-facing endpoints (`get_version_info`, `get_migration_state`)
//! on the SLA calculator are extended with this protocol so that multi-contract
//! backends can verify all contracts agree before deploying.
```

But `get_version_info` in `src/lib.rs` returns `VersionInfo` (5 fields: storage/result schema versions, needs_migration, is_paused, contract_name) — not `VersionNegotiationInfo` (6 fields including `protocol_version` and `min_compatible_protocol`). `negotiate_contract_versions()` and `build_negotiation_info()` are called only by `version_negotiation.rs`'s own tests and the `coordination_harness.rs` test module; no `#[contractimpl]` method exposes them.

Consequences:

- **The documented handshake cannot be performed**: a backend following `docs/VERSION_NEGOTIATION_CONTRIBUTOR_GUIDE.md` cannot obtain the negotiation info from the deployed contract, because no contract method returns it.
- **`protocol_version` exists only in dead code**: the field that drives `NegotiationOutcome::Negotiated`/`Incompatible` is never read from a live contract, so the "staged upgrade" and "incompatible rollback" scenarios in the coordination harness cannot be validated on-chain.
- **Backends that adopt the protocol will build against a phantom API**: the negotiation result type is not reachable, so consumers must either call three separate version endpoints and re-implement the rules themselves (drift risk) or discover the gap at integration time.

## Root cause

The protocol module and its tests were merged (SC-W5-078) without wiring a `#[contractimpl]` entry point; the module doc overstates the integration.

1. Exposing negotiation requires a contract method that accepts peer `VersionNegotiationInfo` values — but that type is `#[contracttype]` and would need to cross the host boundary as an argument, and the peer list must be passed by the caller (the contract cannot call other contracts to fetch their versions without a cross-contract invocation layer).
2. Alternatively, `get_version_info`'s response can be extended with `protocol_version`/`min_compatible_protocol` (append-only additive, safe), giving callers the raw material to run the rules off-chain — but then `negotiate_contract_versions` remains library-only and the "coordinator contract" role is undefined.
3. Any change to `VersionInfo` is an ABI change for existing consumers (`docs/COMPATIBILITY_TRACKING_MATRIX.md`); the append-only rule permits additions, but the doc must state which surface is canonical.

## Acceptance criteria

- [ ] A contract method (or documented combination of existing methods) provides the data needed to run the negotiation rules against a live contract.
- [ ] The module doc in `version_negotiation.rs` and the contributor guide describe the actual integration surface, not an aspirational one.
- [ ] The coordination-harness scenarios are reproducible against the real contract surface (or explicitly marked as library-level tests).

## Out of scope

Building the cross-contract invocation layer (`CrossContractSafety` wiring, companion issue) and the escrow/settlement peer contracts.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/version_negotiation.rs`, `apexchainx_calculator/src/lib.rs` (`get_version_info`), `docs/VERSION_NEGOTIATION_CONTRIBUTOR_GUIDE.md`.

### LINKED ISSUE #428: set_int payload omits mttr_minutes, threshold_minutes, and rating: settlement reconciliation cannot reconstruct the decision from the settlement event
---ISSUE---

## Problem

`publish_settlement_intent_event` (`src/lib.rs`) emits `set_int` with only part of the decision:

```rust
env.events().publish(
    (EVENT_SETTLE_INTENT, EVENT_VERSION, severity),
    (result.outage_id.clone(), result.status.clone(), result.payment_type.clone(),
     result.amount, result.config_version_hash, result.recorded_at),
);
```

The companion `sla_calc` event carries `mttr_minutes`, `threshold_minutes`, and `rating`, but `set_int` does not. The `set_int` docstring calls it a "settlement intent" for "backend reconciliation".

Consequences:

- **Reconciliation consumers must double-read**: a backend reconciling settlements against recorded SLA decisions needs `mttr_minutes`/`threshold_minutes` (to verify the violation/reward math) and `rating` (tier); it must correlate `set_int` back to the `sla_calc` event for the same `outage_id` in the same transaction — an extra join with failure modes when events are processed in isolation.
- **The event is not self-contained despite the doc promise**: the field-order comment says "settlement intent fields are ordered by settlement priority (id, status, payment, amount, hash, timestamp)" — a settlement-only view — so a consumer that processes only `set_int` (the documented "settlement intent" stream) cannot audit the SLA math that produced the amount.
- **`config_version_hash` alone is not enough**: without `threshold_minutes`, a consumer cannot re-derive the violation boundary even with the config registry.

## Root cause

`set_int` was designed as a settlement-handoff payload and dropped the computation fields; the reconciliation use case was documented without the data needed to serve it.

1. Adding fields to `set_int` is an event-schema change: per `docs/EVENT_COMPATIBILITY_POLICY.md` and the SC-099 checklist, appending fields at the end is additive and non-breaking — but the payload's documented ordering ("by settlement priority") must be reconciled with appends.
2. Alternatively `set_int` can be dropped in favor of a single richer `sla_calc` (an ABI removal — breaking, coordinated with indexers), or `dup_input`-style full-result events can become the canonical decision record.
3. The decision interacts with the payload-redundancy companion issue: three events (`sla_calc`, `set_int`, `dup_input`) carry overlapping result fields in different orders; the fix direction should reduce, not extend, the redundancy.

## Acceptance criteria

- [ ] A consumer processing only `set_int` can reconstruct the full SLA decision (status, amount, rating, thresholds, hash).
- [ ] Existing indexers keep working (additive change) or a coordinated migration is documented.
- [ ] Event-schema docs and the SC-099 checklist are updated for the new payload.

## Out of scope

Deduplicating the three overlapping decision events (companion issue) and per-event payload-size guardrails.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`publish_settlement_intent_event`, `publish_sla_event`), `docs/EVENT_COMPATIBILITY_POLICY.md`.

### LINKED ISSUE #429: sla_calc and dup_input order their shared fields differently: indexers must parse two field orders for the same decision data
## Problem

The two decision-carrying events reorder the same fields:

```rust
// sla_calc (src/lib.rs, publish_sla_event)
(outage_id, status, payment_type, rating, mttr_minutes, threshold_minutes, amount)
// dup_input (src/lib.rs, publish_duplicate_input_event)
(outage_id, status, mttr_minutes, threshold_minutes, amount, payment_type, rating, config_version_hash, recorded_at)
```

Both are documented in `src/event_schema.rs` with these divergent layouts.

Consequences:

- **Indexers maintain two decoders**: the `dup_input` event — emitted specifically so consumers can reconcile a rejection "without a second `get_latest_by_outage` read" — encodes `payment_type`/`rating` after `amount`, while `sla_calc` puts them before; a parser written from one schema misreads the other.
- **Field-order drift is invisible to the stability guardrail**: `event_name_symbols()` in `src/api_stability.rs` checks names only, so reordering either payload passes CI (the guardrail cannot detect payload-shape drift at all).
- **The "append-only, additive changes are safe" policy breaks down here**: `dup_input` mirrors `SLAResult`'s struct order while `sla_calc` uses a semantic order — the two events cannot be merged by a consumer without per-event logic.

## Root cause

`sla_calc` predates `SLAResult`'s 9-field struct order; `dup_input` was later written to mirror the struct, and no canonical payload-order rule was enforced across events.

1. Reordering either payload is a breaking ABI change; the realistic fix is to define a canonical field order (e.g. the `SLAResult` struct order) and either migrate `sla_calc` to it with a version bump or document a single shared layout going forward.
2. The SC-099 checklist and `docs/EVENT_DRIFT_CHECKLIST.md` guide reviews but do not mechanically compare payload layouts across events; enforcing a shared order needs a test that asserts the tuple shapes (or generated event-fixture snapshots).
3. `dup_input` deliberately carries two extra fields (`config_version_hash`, `recorded_at`); the canonical order must accommodate the full `SLAResult` shape so all three events align.

## Acceptance criteria

- [ ] A single documented field order applies to the shared fields across `sla_calc`, `set_int`, and `dup_input`.
- [ ] A test (or event-fixture snapshot) fails when a payload tuple is reordered without a version bump.
- [ ] Existing consumers are covered by a migration note when the layout changes.

## Out of scope

Adding the missing fields to `set_int` (companion issue) and payload-size budget enforcement.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (the three `publish_*` helpers), `apexchainx_calculator/src/event_schema.rs`.

JamesEjembi:
Resolves #429,
 Resolves #428,
Resolves  #427,
Resolves  #426.

### #429 — Canonical decision payload across sla_calc / set_int / dup_input
`publish_sla_event` and `publish_settlement_intent_event` now emit the same 9-field canonical tuple `(outage_id, status, mttr_minutes, threshold_minutes, amount, payment_type, rating, config_version_hash, recorded_at)` already used by `dup_input`. A single decoder now parses every decision event. Tests assert the shared canonical order and the 9-field decode.

### #428 — set_int is self-contained for settlement
The settlement-intent event now carries the full decision (rating, mttr minutes, threshold), so settlement-only consumers reconstruct the SLA decision without a follow-up read.

### #427 — get_version_negotiation_info
New read-only method exposing `protocol_version`, `min_compatible_protocol`, storage config version, pause and migration state for the multi-contract handshake. Descriptor entry included; get_public_api reports 62 methods (test updated).

### #426 — accurate address auth labels
`get_public_api` now labels `accept_admin`/`accept_operator` as `"addr"` (they call `require_auth` for the contract address), not `"none"`.

**Verification:** cargo check clean (0 warnings); full test suite 599 passed / 0 failed / 5 ignored (includes regression tests for all four issues).


usmanimamu17-create (approved):
Looks good to me.

usmanimamu17-create (approved):
Looks good, thanks!

PR #546, opened 2026-08-31
Two-step and single-step operator paths coexist with no guidance: set…

### LINKED ISSUE #477: Two-step and single-step operator paths coexist with no guidance: set_operator lets an admin bypass the canonical two-step protocol silently
## Problem

`src/governance.rs` exposes both operator-change protocols, and `get_public_api` lists both:

```rust
// Single-step (admin only, immediate): set_operator  -> op_set
// Two-step (propose -> accept):       propose_operator/accept_operator -> op_prop/op_acc
```

Nothing documents when each is appropriate, and nothing prevents the single-step path from being used in contexts where the two-step protocol was intended (the event schema documents the two-step events as the canonical handoff trail).

Consequences:

- **The audit trail is protocol-dependent**: two equivalent role changes produce different event sequences (`op_set` vs. `op_prop`+`op_acc`); a consumer reconciling "operator changes" must understand both shapes, and an operator change made via `set_operator` has no acceptance step — the new operator never consents (a security-relevant difference: `accept_operator` requires the new operator's signature, `set_operator` only the admin's).
- **The consent gap is invisible in events**: `op_set` carries only `(new_operator,)`; nothing marks that the new operator did not authorize the change.
- **Consumers cannot tell which path was used for policy**: e.g. a backend that alerts when the operator changes without the new operator's consent cannot do so, because the event doesn't distinguish the paths.

## Root cause

`set_operator` (#28-era, single-step) and the later two-step handoff (#64) were both kept for compatibility; the coexistence was never documented or event-tagged.

1. The design decision is whether `set_operator` should be deprecated (removal is an ABI change; the descriptor and any consumers must coordinate), retained as a documented "break-glass" path, or changed to require the new operator's auth (altering its semantics).
2. If both stay, the `op_set` event should carry a marker or the docs must state the consent semantics; event payload changes follow the SC-099 checklist.
3. The security framing matters: `set_operator` without the new operator's consent is a privilege-escalation surface (a compromised admin can install any operator silently), so the issue should weigh deprecation against the break-glass use case.

## Acceptance criteria

- [ ] The two paths' semantics (consent, events, intended use) are documented.
- [ ] Consumers can distinguish the paths from the event stream (or the docs forbid relying on the distinction).
- [ ] Tests cover both paths and the pending-slot interaction (companion issue).

## Out of scope

Removing `set_operator` and the pending-slot orphan (companion issues).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/governance.rs` (`set_operator`, `propose_operator`, `accept_operator`), `apexchainx_calculator/src/event_schema.rs` (op_set/op_prop/op_acc).

### LINKED ISSUE #478: The low-severity penalty cap interacts badly with the cross-severity ordering check: raising low.penalty to 100 forces every higher tier up, contradicting the exemption comment
---ISSUE---

## Problem

`validate_cross_severity_penalty_ordering` (`src/lib.rs`) exempts low from the *upper-direction* check but still enforces the *lower-direction* check against it:

```rust
// The doc says: "low: no cross-severity check (capped independently at 100)"
// ...but the next-lower check still applies when updating medium:
if index + 1 < severities.len() {
    let lower_sev = severities.get(index + 1).ok_or(SLAError::InvalidSeverity)?;
    if let Some(lower_cfg) = configs.get(lower_sev.clone()) {
        if new_penalty < lower_cfg.penalty_per_minute { return Err(SLAError::InvalidPenalty); }
    }
}
```

When the admin sets `low.penalty = 100` (allowed: low's per-severity cap is 100), the next `set_config("medium", ...)` must satisfy `medium.penalty >= 100`, which forces `high >= medium` and `critical >= high` upward through the chain.

Consequences:

- **The documented exemption is illusory**: the comment says low is "capped independently at 100" and exempt, but the ordering rule propagates low's value upward — raising low to 100 effectively forces critical's penalty to at least 100, contradicting the "independent cap" framing.
- **The admin hits a surprise wall later**: setting low to 100 succeeds; the next medium update fails with `InvalidPenalty` unless medium is also raised, and so on — the ordering error surfaces one update late and at the wrong severity.
- **The default config itself sits close to this trap**: defaults are critical=100, high=50, medium=25, low=10 — the progression is strict; any flattening (e.g. low=25) cascades.

## Root cause

The exemption was implemented only for the `index in 1..=2` upper check; the unconditional lower-direction check still treats low as the floor, so the two halves of the rule disagree about low's role.

1. The rule has two directions with different scope (low exempt from "new <= higher", not from "new >= lower"); the correct model is a single monotonicity predicate over the severity order, which requires defining low's position honestly.
2. Fixing it is a config-validation behavior change: either the lower check must also exempt interactions with low (weakening the guarantee), or low's cap must be reconciled with the ladder (changing accepted values).
3. The per-severity caps (critical min 50, high min 25, medium min 10, low max 100) were written as independent bounds; unifying them with the ordering rule into one validated lattice is the real design task, and any change must be pinned by tests for each severity combination.

## Acceptance criteria

- [ ] The ordering rule's behavior for low severity matches its documentation.
- [ ] Tests cover the cascade (low=100 → medium/high/critical updates) and pin the accepted/rejected combinations.
- [ ] The per-severity bounds and the ordering rule are documented as one coherent validation model.

## Out of scope

Threshold ordering (only penalties are ordered) and the custom-severity ordering gap (companion issue in batch 06).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`validate_cross_severity_penalty_ordering`, `validate_config`), `apexchainx_calculator/src/tests.rs` (ordering tests).

### LINKED ISSUE #479: No tests pin the rating-tier boundaries at performance_ratio 50 and 75: compute_result's tier selection at exact boundaries is unspecified by the test suite
## Problem

`compute_result` (`src/lib.rs` and `src/calculation.rs`) selects rating tiers by ratio thresholds:

```rust
let (multiplier, rating) = if performance_ratio < 50 {
    (200u32, symbol_short!("top"))
} else if performance_ratio < 75 {
    (150u32, symbol_short!("excel"))
} else {
    (100u32, symbol_short!("good"))
};
```

A ratio of exactly 50 lands in "excel", exactly 75 in "good". The threshold tests (`src/threshold_config.rs`) cover the met/viol boundary (mttr == threshold) but not the tier boundaries.

Consequences:

- **The tier contract is untested at its edges**: whether `ratio == 50` should be "top" or "excel" is a business decision encoded in a `<` comparison; no test prevents a future change to `<=` from silently moving the boundary — which changes reward amounts (150% vs 200%) for a class of outages.
- **Boundary values are not enumerated**: with a threshold of 30, `mttr == 15` gives ratio 50 (excel) and `mttr == 14` gives 46 (top); a test can pin this, but none exists, so the exact `mttr`/`threshold` combinations that flip tiers are undocumented.
- **The economic-exposure view assumes a specific boundary**: `get_economic_exposure` documents `max_reward` as "matching the `compute_result` path for `performance_ratio < 50`" — the 200% tier — but no test ties the two implementations together (companion issue covers the exposure math).

## Root cause

The tier logic was written as part of the original computation; boundary tests were never added, and the `<` comparisons were never documented as the contract.

1. The fix is tests plus documentation (pin ratio 49/50/74/75 with fixed threshold values), but it surfaces a design question: should the tier thresholds be named constants (e.g. `TOP_RATIO: u64 = 50`) instead of literals, so future tuning is explicit?
2. Tier boundaries interact with the integer-ratio computation (`mttr * 100 / threshold` truncates), so the "true" ratio is an approximation; the tests must pin the truncation behavior at the boundaries, not just the arithmetic.
3. The same boundary question applies to the reward-multiplier constants (200/150/100) used in three places (compute_result, economic exposure, docs) — the fix should decide whether they become shared constants.

## Acceptance criteria

- [ ] Tests pin the tier for ratios 49, 50, 74, 75 (and the mttr/threshold pairs that produce them).
- [ ] The tier thresholds are documented (constants or policy note).
- [ ] `get_economic_exposure`'s 200% assumption matches the tested boundary.

## Out of scope

The 150%-tier rounding behavior (companion issue) and changing the tier values.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/calculation.rs` (`compute_result`), `apexchainx_calculator/src/threshold_config.rs`.

### LINKED ISSUE #486: The 19-entry failure catalog is hardcoded in lib.rs while sl_a_error_count and error_responses.rs mirror it by hand: adding an error requires coordinated edits in four places
## Problem

The error catalog exists in four hand-maintained copies:

```rust
// src/lib.rs, get_failure_schema — a literal 19-entry array
let entries: [(u32, &str, &str); 19] = [
    (1, "AlreadyInitialized", "Contract already initialized"),
    ... (19, "OutageRecalcLimit", "Outage recalc limit reached"),
];

// src/api_stability.rs — pub fn sl_a_error_count() -> u32 { 19 }
// src/error_responses.rs — one is_* helper per variant
// src/lib.rs — the SLAError enum itself
```

Adding error #20 means: extend the enum, extend the literal array, update `sl_a_error_count()`, add an `error_responses.rs` helper, update the stability tests' discriminant list (`[u32; 19]` in `api_stability.rs` tests), and update `docs/FAILURE_TAXONOMY.md`.

Consequences:

- **The catalog is guaranteed to drift**: the enum, the schema array, the count function, and the helper layer are textually independent; a contributor adding a variant can miss any one, and the failure modes are silent (a missing schema entry is invisible to callers unless they enumerate codes).
- **The backend bridge trusts an incomplete schema**: `get_failure_schema` is "the full catalogue" per its doc; if a variant is added without the array entry, backends pre-loading failure codes will not know the error exists — the exact failure the endpoint exists to prevent.
- **The stability count is a second opinion that can disagree**: `sl_a_error_count()` guards the enum length, but nothing ties it to `get_failure_schema`'s array length.

## Root cause

Each layer was written when the enum was smaller and extended by hand; no single source of truth (e.g. deriving the catalog from the enum via a macro or a shared const array) was introduced.

1. The durable fix is to derive the schema from the enum: a macro or a single `const` array of `(code, label, description)` that the enum, the schema, the count, and the helper layer all reference — but `#[contracterror]` enums don't auto-provide descriptions, so the array must be the canonical place and the other layers must consume it.
2. `SLAError` is `#[repr(u32)]` with stable codes; the catalog must preserve the code↔variant mapping (codes are never reused), so any derivation must key off the discriminant.
3. The fix interacts with the stability guardrail (`sl_a_error_count` is asserted in tests) and `docs/FAILURE_TAXONOMY.md`; the change should keep the public behavior identical while removing the duplication.

## Acceptance criteria

- [ ] One source of truth drives the enum count, `get_failure_schema`, and the `error_responses.rs` helpers.
- [ ] Adding an `SLAError` variant without updating the catalog fails a test.
- [ ] `get_failure_schema` is always complete for the deployed enum.

## Out of scope

Adding a new error variant itself and the helper-doc mislabel fix (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`get_failure_schema`, `SLAError`), `apexchainx_calculator/src/api_stability.rs` (`sl_a_error_count`), `apexchainx_calculator/src/error_responses.rs`.

memplethee-lab:
Closes #477
…_operator lets an admin bypass the canonical two-step protocol silently

## Description
Closes #478
## Type of Change
- [ ] Documentation update
Closes #479
## Related Issues

## Changes Made
- 
- 
- 
Closes #486
## Testing

## Checklist


usmanimamu17-create (approved):
Merging this in.

PR #547, opened 2026-08-31
  ## operator transfer paths and add audit-trail tests

### LINKED ISSUE #471: set_operator and propose_operator allow setting the operator to the admin address: role separation collapses silently
## Problem

`src/governance.rs` never checks that the new operator differs from the admin:

```rust
pub fn set_operator(env: &Env, caller: &Address, new_operator: &Address) -> Result<(), SLAError> {
    ... env.storage().instance().set(&OPERATOR_KEY, new_operator); ...
}
pub fn propose_operator(env: &Env, caller: &Address, new_operator: &Address) -> Result<(), SLAError> {
    ... env.storage().instance().set(&PENDING_OP_KEY, new_operator); ...
}
```

Both accept `new_operator == admin` (or `== caller`), collapsing the admin/operator separation the auth model (#28) is built on.

Consequences:

- **A single address can hold both roles**: after `set_operator(admin, admin)`, one key controls config, pause, roles, and calculation — the separation-of-duties guarantee is silently gone, and `get_operator() == get_admin()` is indistinguishable from a legitimate small-deployment choice.
- **The two-party init check is undermined**: `initialize` requires *both* admin and operator to authorize (companion issue) — evidence the design wants distinct roles — yet the runtime can merge them with one admin-only call and no error.
- **Auditors cannot tell intent**: nothing marks an operator==admin state as anomalous; `get_full_audit_state` returns both addresses with no flag.

## Root cause

No invariant check was added when the role setters were written; "admin != operator" was assumed but never enforced.

1. Enforcing the invariant needs a decision on edge cases: is `admin == operator` always invalid, or valid at initialize (which passes both addresses)? If init can create equal roles, the runtime guard must match the init behavior.
2. A rejection is a behavior change for any deployment that deliberately merged roles; the error (new `InvalidInput`-style variant vs. reusing `Unauthorized`) is an ABI decision (companion error-catalog issue).
3. The fix should define the role-separation invariant in one place (a `require_distinct_roles` helper) used by `initialize`, `set_operator`, `propose_operator`, and the admin-transfer path, so all mutation points agree.

## Acceptance criteria

- [ ] Role setters reject (or explicitly flag) an operator equal to the admin.
- [ ] The role-separation invariant is documented and consistent with `initialize`.
- [ ] Tests cover the equal-role rejection for both the single-step and two-step paths.

## Out of scope

Proposing the admin as admin (companion issue) and the initialize two-party auth (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/governance.rs` (`set_operator`, `propose_operator`), `apexchainx_calculator/src/lib.rs` (`initialize`, `require_operator`).

### LINKED ISSUE #473: The config-freeze gate is applied unevenly across admin functions: governance mutations are frozen-gated but pause, unpause, and migrate are not, with no documented boundary
## Problem

`freeze_config` blocks some admin operations but not others:

```rust
// Frozen-gated (src/governance.rs, src/config.rs):
//   propose_admin, accept_admin, cancel_admin_proposal, propose_operator,
//   accept_operator, cancel_operator_proposal, renounce_admin, set_operator,
//   set_config, set_custom_severity, remove_custom_severity
//   (all call config_freeze::require_not_frozen)

// NOT frozen-gated (src/lib.rs / src/metadata.rs):
//   pause, unpause, migrate, set_retention_limit (batch 01 companion)
```

The freeze docs say it "can be used to temporarily prevent configuration changes during critical operations or incident response" — but pause/unpause and migrate are exactly the operations an incident-response lockdown might want to freeze alongside config.

Consequences:

- **The freeze boundary has no stated rationale**: an auditor cannot tell whether the missing gates are intentional (freeze = config-only) or an oversight; the state machine docs (`src/config_freeze.rs`) describe only the config flag, not which operations it governs.
- **A freeze can be circumvented for its stated purpose**: if freeze is meant to lock down during an audit, the admin can still `migrate` (changing storage schema) or `pause`/`unpause` (changing operational state) mid-audit.
- **The gate list must be maintained by hand**: every new admin function must remember to call `require_not_frozen`; nothing structural enforces the boundary (the `get_public_api` descriptor's auth column is similarly manual — companion issue).

## Root cause

`require_not_frozen` was added to the config and governance functions that existed when freeze shipped; later functions (`migrate`, and the pause path was pre-existing) were not audited against the freeze policy, and the policy itself is undocumented.

1. The fix starts with a documented freeze policy: which operations does "config freeze" block? Options include (a) config-only (current, undocumented), (b) all admin mutations, (c) a declared allowlist — each changes behavior for live deployments.
2. If `migrate` becomes frozen-gated, a contract that needs migration *and* is frozen cannot recover without unfreezing first — a deadlock the policy must handle (freeze state is itself changeable, so it resolves, but the ordering matters).
3. `pause`/`unpause` are lifecycle operations with their own semantics; gating them on freeze conflates two state machines unless documented deliberately.

## Acceptance criteria

- [ ] The freeze policy (which operations are blocked) is documented in `src/config_freeze.rs` and `docs/CONTRACT_LIFECYCLE.md`.
- [ ] The implemented gates match the documented policy, with tests for each gated/ungated function.
- [ ] The freeze/unfreeze event stream and the config-freeze state machine agree.

## Out of scope

The `set_retention_limit` freeze gap (companion issue in batch 01) and making the gate structural.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/config_freeze.rs`, `apexchainx_calculator/src/lib.rs` (`pause`, `migrate`), `docs/CONTRACT_LIFECYCLE.md`.

### LINKED ISSUE #480: The 150% reward tier floors odd reward_base values with div_euclid: the rounding policy for non-divisible reward bases is undocumented
## Problem

`compute_result` computes the excel-tier reward as:

```rust
let reward = match cfg.reward_base.checked_mul(multiplier as i128) {
    Some(val) => val.div_euclid(100),
    None => return Err(SLAError::InvalidRewardAmount),
};
```

For `multiplier = 150` and an odd `reward_base` (e.g. 101), `101 * 150 = 15150`, and `15150 / 100 = 151` — the 0.5 is silently floored. The 200% and 100% tiers are exact; only the 150% tier truncates.

Consequences:

- **The reward schedule is not what the formula suggests**: `reward_base * 150 / 100` implies exact arithmetic, but odd bases lose a fractional amount; a backend computing expected rewards by the documented formula (`reward_base * multiplier / 100`) gets `151.5` and must guess the floor.
- **The rounding policy is undocumented and untested**: `div_euclid` truncates toward negative infinity (for positive values, floor); no test asserts the odd-base case, and `docs/` never states whether flooring, rounding, or ceiling is intended.
- **Different severities can pay inconsistent fractions**: two severities with `reward_base` 101 and 100 differ by more than the base difference after tiering (151 vs 150 vs 200 for top), which is fine arithmetically but must be a deliberate policy.

## Root cause

`div_euclid(100)` was chosen for the general case without considering that only the 150% tier produces non-integral products; the policy was never written down.

1. Choosing the rounding policy (floor vs. nearest vs. ceiling) is a business decision affecting every reward payment; changing it is a behavior change with financial consequences and must be versioned.
2. The parity baseline (`test_snapshots/tests/parity_baseline.json`, companion issue) pins reward outputs; any rounding change requires regenerating golden vectors.
3. The same `checked_mul(...).div_euclid(100)` pattern should be extracted so the policy lives in one place (with the multiplier constants from the boundary companion issue).

## Acceptance criteria

- [ ] The rounding policy for non-divisible reward bases is documented (and matches the implementation).
- [ ] Tests pin odd reward_base values across all three tiers.
- [ ] The parity baseline covers an odd-base case.

## Out of scope

The tier-boundary tests (companion issue) and changing the multiplier values.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/calculation.rs` (`compute_result`), `apexchainx_calculator/src/parity_tests.rs`.

### LINKED ISSUE #485: error_responses.rs mislabels is_severity_not_in_set with the OutageRecalcLimit doc comment: the helper layer's docs are copy-pasted
## Problem

`src/error_responses.rs` has a copy-paste doc defect:

```rust
/// Returns `true` if the error is `OutageRecalcLimit`.
pub fn is_severity_not_in_set(err: &SLAError) -> bool {
    matches!(err, SLAError::SeverityNotInSet)
}
```

The function checks `SeverityNotInSet` (code 18) but is documented as checking `OutageRecalcLimit` (code 19). The real `OutageRecalcLimit` helper (`is_outage_recalc_limit`) is correctly labeled on the line after.

Consequences:

- **The helper layer teaches the wrong mapping**: a backend developer using `error_responses.rs` as the dispatch guide (its stated purpose: "error-type dispatch without matching on the raw enum discriminant") will read that `SeverityNotInSet` is the recalc-limit error and wire the wrong handler.
- **The two adjacent helpers are easy to mis-swap**: `is_severity_not_in_set` and `is_outage_recalc_limit` sit next to each other with near-identical docs; the mislabeled one is a trap for exactly the copy-paste it came from.
- **The doc/behavior mismatch is untested**: no test asserts each helper matches its namesake variant (the batch 01 companion issue on the failure catalog covers the same drift class).

## Root cause

The helper was added by copying the `OutageRecalcLimit` doc block and editing only the body; the doc line was left behind.

1. The fix is a one-line doc correction, but the durable fix is a test that pairs each `is_*` helper with its `SLAError` variant (the #264-style "failure catalog drift" test the repo's closed issues reference) — without it, the next helper added will drift again.
2. `error_responses.rs` is the backend bridge layer; its correctness is a consumer contract, so a doc fix should be accompanied by a check that every helper's docstring names the variant it matches.
3. This is a good candidate for a lint/comment test that scans the file for `is_*` functions whose docstring variant name differs from the `matches!` target.

## Acceptance criteria

- [ ] `is_severity_not_in_set`'s doc names `SeverityNotInSet`.
- [ ] A test (or lint) fails when a helper's docstring names a different variant than its `matches!` target.
- [ ] All 19 helpers' docs are verified against their bodies.

## Out of scope

The failure-catalog single source of truth (companion issue) and adding new helpers.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/error_responses.rs`, `apexchainx_calculator/src/lib.rs` (`SLAError`).

akargi:
Closes #471
## Description

## Type of Change
- [ ] Documentation update
Closes #473
## Related Issues
Closes #480
## Changes Made
- 
- 
- 

## Testing

## Checklist

Closes #485


usmanimamu17-create (approved):
LGTM.

PR #548, opened 2026-08-31
Feat/dev branch

### LINKED ISSUE #481: get_economic_exposure totals use saturating_add with no signal: unlike increment_stats, exposure overflow caps silently
## Problem

`get_economic_exposure` (`src/lib.rs`) aggregates totals with silent saturation:

```rust
total_max_reward = total_max_reward.saturating_add(max_reward);
total_penalty_per_minute = total_penalty_per_minute.saturating_add(penalty_rate);
```

The running stats path (`increment_stats`) went to lengths to surface saturation — `checked_add` + `stats_sat` events (SC-W5-047) — but the exposure view caps at `i128::MAX` with no signal.

Consequences:

- **The exposure view can under-report with no indication**: with four canonical severities plus custom entries (if the companion issue lands), sums near `i128::MAX` clamp silently; a dashboard reading `total_max_reward == i128::MAX` cannot tell whether that is a real configured total or a saturation artifact.
- **The two aggregation paths set different precedents**: `increment_stats` treats saturation as reportable (event, capped value), `get_economic_exposure` treats it as invisible — a consumer that learned the stats convention will not expect the view to lie silently.
- **The view is read-only and stateless**: it cannot emit events, so the fix must either avoid saturation (e.g. return an error) or expose the saturation state in the response.

## Root cause

The view was written with `saturating_add` as a cheap overflow guard, predating the `stats_sat` convention; the two paths were never reconciled.

1. The response types (`EconomicExposure`, `SeverityExposure`) are `#[contracttype]` and stability-guarded; adding a `saturated: bool` field is additive but requires the guardrail update (companion issue).
2. Returning an error on overflow changes the endpoint's contract (read-only views are expected to succeed); the design decision is signal-in-response vs. error.
3. In practice the configured bounds (reward_base ≤ 100,000, four severities) make overflow unreachable today — so the issue is about *correctness-by-construction*: either document the unreachability or make the math checked so a future bound change cannot silently corrupt totals.

## Acceptance criteria

- [ ] Exposure totals cannot silently cap (checked math with a defined signal, or documented unreachability with an assertion).
- [ ] The behavior is consistent with the `stats_sat` convention or explicitly divergent with rationale.
- [ ] Tests cover the extreme-configuration case.

## Out of scope

The custom-severity omission in the view (companion issue in batch 06) and the reward-tier boundary tests.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`get_economic_exposure`, `increment_stats`).

### LINKED ISSUE #482: The weekly telemetry reset freezes forever once the ledger timestamp passes u32::MAX: the clamp makes last-calc permanent and stale detection never fires again
## Problem

`record_severity_telemetry` (`src/lib.rs`/`src/calculation.rs`) stores "last calculation" as a `u32` lane and clamps when the timestamp exceeds `u32::MAX`:

```rust
let current_ledger = if now > u64::from(u32::MAX) { u32::MAX } else { now as u32 };
last_calculations = Self::set_count_lane(last_calculations, index, current_ledger);
```

The weekly-reset check then compares `now.saturating_sub(last_calc) >= week_seconds`. Once `now` exceeds `u32::MAX` seconds (~year 2106) and the lane is clamped to `u32::MAX`, `now - u32::MAX` stays below a week forever — the lane is never stale, and the weekly reset never fires again.

Consequences:

- **Telemetry windows silently become cumulative after 2106**: the reset logic is dead code from that point; counters grow monotonically and `violation_rate` drifts toward the cumulative value.
- **The truncation is a latent overflow bug**: `now as u32` would wrap for any timestamp above `u32::MAX` if the clamp were removed; the clamp trades a wrap for a permanent freeze, neither of which is documented.
- **The field name is wrong at the storage level**: `LAST_CALCULATION_LEDGER_KEY` stores a timestamp (batch 01 companion issue), and the u32 width is the root of this edge — the two issues share a root cause.

## Root cause

The timestamp was squeezed into a `u32` lane (shared with the count packing scheme) and the clamp was added without considering that it breaks the staleness predicate.

1. The fix is a storage-format change: the `u32` lane cannot hold timestamps past 2106; moving to `u64` per severity breaks the packed-`u128` lane scheme (or requires two `u128` keys), which is a storage-schema change subject to the migration checklist.
2. The weekly window itself is a design choice (7 days); the fix should define what "stale" means robustly (e.g. compare against a monotonic ledger sequence instead of wall time — the batch 01 companion issue makes this argument).
3. Tests run at synthetic timestamps; a test can set `now > u32::MAX` in the Env to pin the freeze — the issue should require that test to exist.

## Acceptance criteria

- [ ] The weekly reset still fires for timestamps beyond `u32::MAX` (or the windowing scheme is redesigned to be overflow-proof).
- [ ] A test exercises the post-2106 timestamp path and pins the behavior.
- [ ] The telemetry storage format is documented with its range limits.

## Out of scope

The timestamp-vs-ledger naming (batch 01 companion) and custom-severity lane misattribution (batch 01 companion).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`record_severity_telemetry`), `apexchainx_calculator/src/calculation.rs` (same function).

### LINKED ISSUE #483: No-op set_config with identical values still bumps the config version hash: duplicate-protection generations and LCFGUPD advance without any real change
## Problem

`set_config` (`src/lib.rs`/`src/config.rs`) writes, re-hashes, and stamps the update sequence even when the values are unchanged:

```rust
configs.set(severity.clone(), SLAConfig { threshold_minutes, penalty_per_minute, reward_base });
env.storage().instance().set(&CONFIG_KEY, &configs);
config_metadata::record_config_update(&env);   // LCFGUPD = ledger sequence
env.events().publish((EVENT_CONFIG_UPD, EVENT_VERSION, severity), (...));
```

A call that sets the exact current values changes `compute_config_version_hash` (the polynomial includes the values, and the hash is re-derived — but if values are identical, the hash is identical... unless the write reorders nothing; the hash would be the same). The real effects: `LCFGUPD` advances (a new sequence is stamped) and the `cfg_upd` event fires, so *consumers* see a config update while the hash is unchanged.

Consequences:

- **Cache invalidation fires without a config change**: backends that poll `get_last_config_update` (the documented invalidation signal, #4) will re-fetch the full config after a no-op write and find nothing different — wasted reads and, worse, a signal they cannot distinguish from a real change.
- **The event stream over-reports config activity**: `cfg_upd` fires for a no-op; event-based reconstruction (companion issue) records a phantom generation.
- **Duplicate-detection generations advance on no-ops**: `calculate_sla` keys recalculation eligibility on `config_version_hash`; since the hash does not change on a no-op, existing entries are not invalidated — but consumers cannot tell from the update stamp that nothing changed, so they may treat the (unchanged) hash as "changed" and expect fresh generations that never come.

## Root cause

`set_config` never compares the new values to the stored ones; the write/event/stamp path treats every call as a change.

1. The fix is an early-return on value equality (compare the `SLAConfig` before writing) — but "no change" must be defined exactly (field equality on the stored config for that severity), and the return must skip the event, the `LCFGUPD` stamp, and the storage write.
2. Skipping the event is a behavior change for consumers that may rely on every `set_config` call emitting (e.g. counting calls); the issue must decide whether a no-op is truly silent or emits a distinct "no-op" signal.
3. The batch 01 companion issue notes custom-severity changes never touch the canonical hash — the no-op question and the hash-exclusion question must be resolved together so "config changed" has one definition across all paths.

## Acceptance criteria

- [ ] A no-op `set_config` does not stamp a new update sequence or emit a change event (or emits a clearly distinct one).
- [ ] `get_last_config_update`/`get_config_version_hash` agree on whether a change occurred.
- [ ] Tests cover no-op writes for each severity (canonical and custom).

## Out of scope

The custom-severity hash exclusion (companion issue in batch 01) and the config-catalog single source of truth.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`set_config`), `apexchainx_calculator/src/config_metadata.rs`.

### LINKED ISSUE #484: compute_config_version_hash is a 63-bit polynomial with no collision analysis: two colliding configs defeat duplicate detection silently
## Problem

`compute_config_version_hash` (`src/lib.rs`) hashes the 12 config values into a `u64`:

```rust
const BASE: u64 = 91138233;
const MODULUS: u64 = (1u64 << 63) - 25;
let mut hash: u64 = 1;
... // polynomial over threshold/penalty/reward for each severity
```

The hash drives duplicate detection (`DuplicateOutageInput` vs. replay vs. fresh generation) and backend cache invalidation. There is no collision analysis, no test that searches for colliding configs, and no documented bound on the collision probability.

Consequences:

- **A collision silently disables duplicate protection**: two distinct configs with the same hash mean an outage resubmitted under config B is treated as "unchanged config" vs. the entry recorded under config A — the `dup_input`/replay/recalc state machine (SC-W5-046) misclassifies the submission, and the contract stores or rejects based on the wrong generation.
- **The failure is silent and permanent**: unlike a hash-collision in a Merkle structure (detectable on comparison), here the hash is the *only* comparison key; a collision produces wrong behavior with no error.
- **The modulus `(1<<63)-25` is unusual**: it is not a known prime (the doc calls the scheme a "polynomial rolling hash with a prime base and modulus" — the modulus's primality is unverified), and no analysis documents why this construction was chosen.

## Root cause

The hash was implemented as a deterministic fingerprint without an adversarial/collision assessment; the state machine was then built to trust it absolutely.

1. The search space is bounded (threshold 1..1440, penalty 1..10000, reward 1..100000 per severity) — roughly 10^15 combinations, too large for exhaustive search but the polynomial's structure can be analyzed for algebraic collisions (the same config values in different severity slots interact through the shared `power` schedule).
2. The fix options are: (a) prove/bound the collision probability for the constrained value space, (b) switch to a stronger construction (e.g. two independent hashes or a cryptographically-influenced mix), or (c) store the full config (or its canonical serialization) as the comparison key — a storage change since `SLAResult.config_version_hash` is a `u64` field (ABI-guarded).
3. Any change to the hash breaks the equivalence between historical results' stored hashes and newly computed ones — existing history entries carry the old hash, so a hash-scheme change needs a migration strategy (or a versioned hash with both computed).

## Acceptance criteria

- [ ] The collision probability is analyzed and documented for the constrained value space (or the construction is replaced with an analyzed one).
- [ ] The state machine's trust in the hash is documented (what a collision would mean for each path).
- [ ] Tests search a meaningful subset of the value space for collisions.

## Out of scope

The custom-severity hash exclusion (companion issue) and replacing `config_version_hash` with a stored config reference.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`compute_config_version_hash`), `apexchainx_calculator/src/tests.rs` (duplicate-detection tests).

DevMuhdishaq:
Closes #481
## Problem

`get_economic_exposure` aggregated per-severity `max_reward` and
`penalty_per_minute` totals using `saturating_add` and
`checked_mul(…).unwrap_or(i128::MAX)`, which silently capped at
`i128::MAX` with no signal to consumers. A dashboard reading
`total_max_reward == i128::MAX` could not distinguish a real configured
total from a saturation artifact.
Closes #482
Meanwhile, `increment_stats` (SC-W5-047) uses `checked_add` and emits
`stats_sat` events on overflow — two aggregation paths with
inconsistent overflow semantics.

## Solution

Replace all silent saturation in `get_economic_exposure` with checked
arithmetic that returns a new `SLAError::ExposureOverflow` (code 22):
Closes #483
- `checked_mul(200).unwrap_or(i128::MAX)` → `.ok_or(SLAError::ExposureOverflow)?`
- `saturating_add(max_reward)` → `checked_add(…).ok_or(SLAError::ExposureOverflow)?`
- `saturating_add(penalty_rate)` → `checked_add(…).ok_or(SLAError::ExposureOverflow)?`

The error is chosen over a `stats_sat`-style event because the view is
read-only and stateless — it cannot emit events. The view's return type
(`EconomicExposure`) is `#[contracttype]` and stability-guarded, so
adding a `saturated: bool` field was deferred to a companion issue.
Returning a typed error gives consumers a clear, machine-dispatchable
signal without changing the response schema.

### Why this is unreachable today

With current validation bounds (`reward_base ≤ 100 000`,
`penalty_per_minute ≤ 10 000`, 4 canonical severities) the maximum
totals are `800 000` and `40 000` — orders of magnitude below
`i128::MAX`. Checked arithmetic is used regardless for
correctness-by-construction: a future bound relaxation or
custom-severity expansion will produce a clear error instead of silently
corrupting dashboard totals.
Closes #484
## Files changed

| File | Change |
|------|--------|
| `apexchainx_calculator/src/lib.rs` | Add `ExposureOverflow = 22` error variant; replace `saturating_add` / `unwrap_or` with checked arithmetic in `get_economic_exposure`; add error docs; update `get_failure_schema` entries (21→22) |
| `apexchainx_calculator/src/error_responses.rs` | Add `is_exposure_overflow` helper |
| `apexchainx_calculator/src/tests.rs` | Add two overflow tests (reward + penalty); update `is_terminal`, error classification counts (19→22), and `test_failure_catalog_matches_error_helpers` |
| `docs/FAILURE_TAXONOMY.md` | Add Code 22 taxonomy entry and summary table row |
| `docs/sla-error-additions-guide.md` | Add codes 20, 21, 22 to the error code table |

## Tests

- `test_economic_exposure_returns_error_on_overflow` — injects extreme `reward_base` via raw storage, asserts `ExposureOverflow`
- `test_economic_exposure_returns_error_on_penalty_overflow` — same for `penalty_per_minute`
- All 600 existing tests pass; clippy clean


usmanimamu17-create (approved):
Approved, let's go.

PR #549, opened 2026-08-31
feat: complete storage and audit hardening

### LINKED ISSUE #463: get_full_audit_state re-reads the full history to report history_len: a bootstrap call materializes up to 1000 entries just to count them
## Problem

`get_full_audit_state` (`src/lib.rs`) fetches the entire history vector and then reports only its length:

```rust
let history: Vec<SLAResult> = env.storage().instance().get(&HISTORY_KEY).unwrap_or_else(|| Vec::new(&env));
let history_len = history.len();
Ok(AuditState { ... history_len, ... })
```

The method is documented as "Combined audit-state envelope for one-shot backend bootstrap reads" (#107) — a call meant to be cheap and one-shot.

Consequences:

- **The one-shot bootstrap is the most expensive read**: it already makes 9+ storage reads (admin, operator, pending, pause, config, stats, schema); adding a full 1000-entry history deserialization just to count makes the "one RPC bootstrap" pay the largest read in the contract.
- **The returned `history_len` and `get_history` can disagree on cost but not value**: consumers that need the count get it at full-vector cost, and consumers that then page history pay it again — double materialization of the same data.
- **The envelope's own doc ("one-shot bootstrap") is at odds with its cost**: the whole point of the envelope is to avoid N round-trips; the history length is the one field that reintroduces a heavy read.

## Root cause

`AuditState` includes `history_len`, and the simplest implementation read the whole vector; no length-only accessor exists (instance storage is whole-value, so even `len()` needs the deserialization unless the length is stored separately).

1. A cheap `history_len` requires storing the length separately (a counter key maintained on append/trim/prune — new storage keys and migration) or reading the vector anyway (current behavior); the redesign decision belongs with the history-storage companion issue.
2. Alternatively, `AuditState` can drop `history_len` (an ABI change to a `#[contracttype]` struct — stability-guarded) and consumers call `get_history_page_with_meta(0, 1)` whose `total` field gives the length from a read that can stop after the slice... but it still deserializes the full Vec today.
3. The envelope is `#[contracttype]` and stability-guarded (`canonical_field_counts` omits it today — companion issue); any shape change must follow the guardrail process.

## Acceptance criteria

- [ ] Obtaining `history_len` does not require materializing the full history (counter or length-aware storage).
- [ ] The bootstrap envelope's cost is documented and measured.
- [ ] `AuditState` shape changes (if any) follow the stability process.

## Out of scope

The full history-storage redesign (companion issue) and the accessor materialization (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`get_full_audit_state`), `apexchainx_calculator/src/audit_state.rs`.

### LINKED ISSUE #464: get_history_page and get_history_page_with_meta duplicate the slicing logic: pagination semantics live in two places and can diverge
## Problem

The two paginated accessors implement the same policy twice:

```rust
// get_history_page (src/lib.rs and src/history.rs — both copies)
if offset >= len || limit == 0 { return Ok(page); }
let end = offset.saturating_add(limit).min(len);
for i in offset..end { page.push_back(history.get(i).unwrap()); }

// get_history_page_with_meta — identical saturating/min logic plus has_more
let end = offset.saturating_add(limit).min(total);
if offset < total && limit != 0 { for i in offset..end { items.push_back(...); } }
Ok(HistoryPage { items, total, has_more: end < total })
```

Each file pair (lib.rs and history.rs) duplicates it again — the pagination policy (issue #263) exists as four copies of the same ~10 lines, all with the same explanatory comment.

Consequences:

- **The policy can drift between copies**: a fix to one (e.g. clamping `limit`) without the others leaves the four accessors with different behavior for the same documented policy — the exact failure the policy doc was written to prevent.
- **`items` and `has_more` can disagree**: `get_history_page_with_meta`'s doc promises `items` is "identical to what `get_history_page` returns for the same (offset, limit)"; that equivalence is enforced by duplication, not by sharing.
- **Review of a pagination change is a 4-site diff**: any contributor fixing a pagination edge must find and update all copies.

## Root cause

The pagination work (#263) was applied by copying the block into each accessor (and each module copy) rather than extracting a shared helper.

1. The fix is a shared `fn page_slice(history, offset, limit) -> (Vec<SLAResult>, end_index)` helper used by both accessors — mechanically simple, but the two modules (`lib.rs`, `history.rs`) each carry full copies of the accessors (the module-duplication companion issue), so the helper must be placed where both can reach it without expanding the duplication.
2. Behavior must be pinned first: the policy doc and tests define the semantics (empty page, `has_more`, saturation); the refactor must be covered by the existing pagination tests before consolidation.
3. The `history.rs` copy exists because `lib.rs` re-exposes the same methods; the right long-term shape depends on the module-ownership decision (which module owns history).

## Acceptance criteria

- [ ] A single implementation of the pagination slice is used by both accessors.
- [ ] The documented equivalence (`items` identical, `has_more` correct) is asserted by a test.
- [ ] Existing pagination policy tests pass unchanged.

## Out of scope

The history storage redesign (companion issue) and the module-duplication cleanup (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/history.rs`, `apexchainx_calculator/src/lib.rs` (`get_history_page*`), `docs/HISTORY_PAGINATION_POLICY.md`.

### LINKED ISSUE #465: compute_result documents recorded_at as "0 in view/audit mode" but calculate_sla_view passes the live ledger timestamp: audit-mode timestamp semantics are documented two conflicting ways
## Problem

Two docstrings describe `recorded_at` differently:

```rust
// src/lib.rs, compute_result (private, in lib.rs and public in calculation.rs)
/// `config_version_hash` binds the result to the exact config snapshot used
/// during evaluation. `recorded_at` is the ledger timestamp at call time
/// (0 in view/audit mode).

// src/lib.rs, calculate_sla_view — the actual call site
Self::compute_result(outage_id, mttr_minutes, &cfg, config_version_hash, env.ledger().timestamp())
```

`calculate_sla_view` passes the live timestamp, so the "0 in view/audit mode" comment is false. `src/history.rs`'s `prune_history_by_age` docstring repeats the same fiction: "Entries with `recorded_at == 0` (view-mode results that were never stored with a real timestamp) are always kept" — but view results are never stored at all, so the `== 0` carve-out has no real producer (every stored result comes from `calculate_sla`, which passes a live timestamp).

Consequences:

- **The audit-mode timestamp contract is ambiguous**: a consumer deciding whether view results carry a meaningful `recorded_at` gets two answers from the code's own docs.
- **The prune carve-out is dead behavior with misleading docs**: `prune_history_by_age` special-cases `recorded_at == 0` for a case that cannot occur through the mutating path; a future change that does store zero timestamps (e.g. bulk import) would silently flip prune semantics.
- **The contradiction obscures a real question**: should `calculate_sla_view` stamp the evaluation timestamp (current behavior, useful for "what would the result be now") or `0` (as documented)? The answer determines whether view results can be compared to stored results by `recorded_at`.

## Root cause

The `compute_result` docstring was written when the view path used `0`; the call site was later changed to pass the live timestamp, and neither the docstring nor the prune doc was updated.

1. Choosing the correct semantic is a design decision with downstream effects: `recorded_at` drives `prune_history_by_age` cutoffs and any consumer that buckets results by time; view results (never stored) only matter if consumers compare view output to stored history.
2. The fix is primarily documentation + a test pinning `calculate_sla_view`'s `recorded_at` behavior, but it should also decide whether the `== 0` carve-out in `prune_history_by_age` remains meaningful.
3. This is a good companion to the audit-path companion issues (view emits no events, replay uses current config) — the "audit mode" semantics deserve one coherent statement.

## Acceptance criteria

- [ ] The `compute_result` and `prune_history_by_age` docs describe the actual `recorded_at` semantics.
- [ ] A test pins `calculate_sla_view`'s `recorded_at` value.
- [ ] The audit-mode behavior (timestamps, events, storage) is documented in one place.

## Out of scope

Changing `calculate_sla_view` to emit events or store results (companion issues).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`compute_result`, `calculate_sla_view`), `apexchainx_calculator/src/history.rs` (`prune_history_by_age`).

### LINKED ISSUE #466: No CI budget assertion ties prune_benchmark and storage_footprint_tests to calculate_sla's per-call storage write cost: write-amplification regressions go unflagged
## Problem

The repo has performance-aware test modules — `prune_benchmark.rs` and `storage_footprint_tests.rs` — but nothing asserts a budget:

- `prune_benchmark.rs` measures prune operations; `storage_footprint_tests.rs` measures footprint growth.
- CI (`ci.yml`) runs `cargo test --lib` (which includes them) and separate `schema_migration_tests`; no job compares measurements to thresholds.
- `calculate_sla` rewrites the entire history `Vec` on every submission (companion issue); nothing fails if that per-call write cost doubles.

Consequences:

- **The benchmark is a report, not a gate**: the write-amplification cost is the single most predictable per-call expense (O(retained) storage write on every `calculate_sla`), yet a regression (e.g. history stored twice, extra deserialization) passes CI silently as long as tests still pass.
- **The footprint tests assert only growth direction**: `assert!(updated_footprint > initial_footprint)`-style checks verify monotonicity, not magnitude — a change that triples per-entry cost is green.
- **The performance story is anecdotal**: `docs/PRUNING_BENCHMARK_NOTE.md` and `docs/BENCHMARK_SEVERITY_COSTS.md` exist, but no CI artifact ties the numbers to a budget.

## Root cause

The benchmark modules were added to *measure*, and the budgets were never defined or enforced; CI runs tests for correctness, not cost.

1. Enforcing a budget requires stable, comparable measurements: the benchmark must run in a controlled Env (deterministic ledger, fixed history size) and assert a byte/op threshold — thresholds can flake on host variance, so the design must pick robust metrics (e.g. history-vec serialized size, not wall time).
2. The budget is a contract decision: what per-call write cost is acceptable at `MAX_HISTORY_SIZE = 1000`? Setting the number is a product/maintenance decision the issue must surface.
3. The measurement and the storage redesign (companion issue) interact: if history storage changes, the budgets must be re-baselined — the issue should define how budgets are versioned.

## Acceptance criteria

- [ ] A CI gate asserts a per-call write-cost budget for `calculate_sla` at a pinned history size.
- [ ] The benchmark outputs are comparable across runs (deterministic metric).
- [ ] Budget changes are documented and reviewed like ABI changes.

## Out of scope

The storage redesign itself (companion issue) and the footprint estimate calibration (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/prune_benchmark.rs`, `apexchainx_calculator/src/storage_footprint_tests.rs`, `apexchainx_calculator/src/lib.rs` (`calculate_sla`).

Agaki00:
Closes #463 , 
Closes #464 , 
Closes #465 and 
Closes #466

---

## Summary\n- Backfill cached history length through storage migration v1 to v2 (#463).\n- Keep pagination accessors aligned through a shared slice implementation (#464).\n- Pin live-timestamp semantics for calculate_sla_view (#465).\n- Add high-history write-cost and bootstrap-read regression gates (#466).\n\n## Verification\n- cargo fmt --check\n- cargo check --lib\n- targeted migration, fingerprint, metadata, pagination, and timestamp tests\n\nThe full suite has expensive storage-cost coverage and generated local snapshots that are intentionally not included in this PR.


usmanimamu17-create (approved):
Looks great, approved.

usmanimamu17-create (approved):
Solid work, merging 🚀

PR #550, opened 2026-08-31
fix: close #491, #492, #493, #494 — orphan lint, api-descriptor guardrail, pause_info convention, fingerprint error propagation

### LINKED ISSUE #491: scratch_test.rs is committed with a test_sym helper that is never declared as a module: leftover scratch code ships in the crate
## Problem

`apexchainx_calculator/scratch_test.rs` exists at the crate root with:

```rust
use soroban_sdk::{Env, Symbol};

pub fn test_sym(env: &Env) -> Symbol {
    Symbol::new(env, "migrate_done")
}
```

It is never declared as a module (`src/lib.rs` has no `mod scratch_test;`), so it is not compiled, its "test" runs nothing, and no tooling flags it (the orphan-file companion issue).

Consequences:

- **Scratch code is version-controlled**: the file is a leftover interactive experiment; its presence signals to contributors that scratch files in the crate are acceptable.
- **The helper it contains is a trap**: `test_sym` appears to be a quick experiment for the `migrate_done` symbol (which the real code builds via `Symbol::new(&env, event_schema::EVENT_MIGRATE_DONE)`); a contributor might import from scratch_test.rs and get an unmaintained duplicate.
- **Orphan detection has no teeth**: CI (cargo check/clippy/tests) ignores the file entirely, so the scratch pattern can accumulate.

## Root cause

The file was committed during the migration-event work and never cleaned up.

1. The fix is deletion, but the general problem — orphan `.rs` files invisible to the compiler — needs the CI lint proposed in the orphan-module companion issue (compare `src/**/*.rs` against declared modules); without it, the next scratch file returns.
2. The repo has a `CODING_STYLE.md` and contributor docs; a note on where experiments belong (e.g. `/tmp` or the fuzz corpus) should accompany the cleanup to prevent recurrence.
3. This issue is deliberately small — a good first contributor task — but must be resolved in the same release as the orphan lint so the class is closed, not just the instance.

## Acceptance criteria

- [ ] `scratch_test.rs` is removed.
- [ ] A lint/test fails when a `.rs` file under the crate is not declared as a module.
- [ ] Contributor docs state where throwaway code belongs.

## Out of scope

The other orphan modules (`storage_version.rs`, `deployment_policy.rs`, `payload_optimizer.rs`, `policy.rs`, `threshold_config.rs`) — companion issue.

## Getting started

```bash
rm apexchainx_calculator/scratch_test.rs
just test
```

Good first files to read: `apexchainx_calculator/scratch_test.rs`, `apexchainx_calculator/src/lib.rs` (module list).

### LINKED ISSUE #492: No test asserts get_public_api matches the #[contractimpl] surface: the descriptor drift is exactly what such a test would catch
## Problem

`get_public_api` (`src/lib.rs`) hand-enumerates the API surface (companion issue documents the drift: `get_contract_info`, `get_storage_footprint_estimate`, and `get_rent_estimate` are missing). No test or CI step compares the descriptor to the actual `#[contractimpl]` methods, so the drift shipped without any red flag.

Consequences:

- **The drift class is unguarded**: the descriptor's documented purpose is to let backends "validate that the deployed contract matches their expected API surface", but the descriptor itself has no validator; the most important consumer of the descriptor cannot trust it.
- **Every future method addition re-introduces the risk**: a contributor adding a public method and updating the descriptor's list has no safety net if they forget (or miscount) — the failure only appears when a backend queries for the new method.
- **The `auth`/`event`/`mutates` columns drift the same way**: even if the method list were complete, the metadata per method is unchecked (the `initialize` auth mislabel, companion issue).

## Root cause

The descriptor was written as static data with no generating or verifying mechanism; the stability guardrails (`api_stability.rs`) check struct field counts, not method surfaces.

1. Verifying "the descriptor matches the impl" requires enumerating the `#[contractimpl]` methods at test time — feasible via the generated client (the SDK exposes the contract's methods through `SLACalculatorContractClient`), so a test can iterate the client's methods and assert the descriptor covers them.
2. The `auth`/`mutates` columns cannot be derived mechanically from the Rust signatures today (they encode intent), so the test can only enforce completeness of names, not correctness of metadata — the metadata needs its own review path.
3. The append-only stability rule for `PublicApiDescriptor` means the test must distinguish "descriptor missing a method" (failure) from "descriptor has a method the impl dropped" (also a failure — the impl is the truth).

## Acceptance criteria

- [ ] A test fails when a public `#[contractimpl]` method is absent from `get_public_api`.
- [ ] A test fails when the descriptor lists a method the impl no longer exposes.
- [ ] The test is documented as the enforcement mechanism for descriptor completeness.

## Out of scope

Deriving `auth`/`mutates` mechanically and fixing the current drift (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`get_public_api`), `apexchainx_calculator/src/tests.rs` (client usage patterns).

### LINKED ISSUE #493: The pause_info Option workaround is applied inconsistently: AuditState exposes Vec<PauseInfo> while other surfaces use Option<Address>
## Problem

The `#[contracttype]`-can't-do-`Option` limitation is worked around inconsistently across the public surface:

```rust
// src/audit_state.rs — pause_info is a Vec<PauseInfo> with a comment:
// "A `Vec` stands in for `Option` because `#[contracttype]` cannot convert Option<PauseInfo>."
pub pause_info: Vec<PauseInfo>,

// src/lib.rs — get_pending_admin / get_pending_operator return Option<Address> directly,
// and get_pause_info returns Result<Option<PauseInfo>, SLAError>
pub fn get_pending_admin(env: Env) -> Result<Option<Address>, SLAError>
```

Consumers of `get_full_audit_state` must know the `Vec` convention (empty = not paused), while consumers of `get_pause_info`/`get_pending_admin` use `Option` — two conventions for "absent" in the same API family.

Consequences:

- **The same semantic (absent pause metadata) has two shapes**: a backend that switches from `get_pause_info` (Option) to `get_full_audit_state` (Vec) must special-case the representation; a consumer that treats any non-empty Vec as "paused with multiple incidents" would be wrong (the Vec is documented as max one element).
- **The workaround comment is a code smell reviewers now expect everywhere**: new `#[contracttype]` structs will re-discover the limitation (the `ConfigBundle` issue #1 was the same trap), and there is no central guidance on the preferred convention.
- **`pause_info: Vec<PauseInfo>` allows an invalid state**: nothing enforces the documented "single-element when paused" invariant; a future change could push multiple entries and consumers would misread it.

## Root cause

Each surface was written when the SDK limitation was hit; the conventions were never standardized or documented centrally.

1. Standardizing requires either a shared wrapper type (e.g. `OptionLike<T>` — new ABI types, stability-guarded) or a documented rule (e.g. "`#[contracttype]` surfaces use Vec with max-length invariants; getters use Option") that all modules follow.
2. Changing `AuditState.pause_info` to a different shape is an ABI change for a `#[contracttype]` struct (the guardrail omits it today — companion issue), so any change must go through the stability process.
3. The deeper fix is upgrading the SDK version if a newer SDK handles `Option` in contract types — a dependency decision with broad blast radius; until then the convention must be written down.

## Acceptance criteria

- [ ] One documented convention governs "absent" state across `#[contracttype]` surfaces.
- [ ] The `AuditState.pause_info` invariant (empty or single-element) is enforced or documented.
- [ ] New `#[contracttype]` structs follow the documented convention (contributor doc update).

## Out of scope

SDK upgrades and the `ConfigBundle`-style deserialization traps (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/audit_state.rs`, `apexchainx_calculator/src/lib.rs` (`get_pause_info`, `get_pending_admin`).

### LINKED ISSUE #494: ContractStateFingerprint silently reports config_version_hash 0 via unwrap_or when the config is unreadable: a corrupt config is indistinguishable from a legitimate zero hash
## Problem

`get_contract_state_fingerprint` (`src/lib.rs`) swallows config-hash failures:

```rust
let config_version_hash: u64 = Self::compute_config_version_hash(&env).unwrap_or(0);
```

The fallback value `0` is also the polynomial's seed behavior — a legitimately computed hash is never 0 (the seed is 1 and the schedule multiplies/adds), so `0` is only reachable via this fallback. The function then returns `Ok(fingerprint)` with `config_version_hash: 0`.

Consequences:

- **A corrupt/unreadable config looks like a valid state**: if `CONFIG_KEY` is unreadable (storage corruption, a bad migration), the fingerprint reports `config_version_hash = 0` with `Ok` — an operator comparing pre/post-upgrade fingerprints (the documented use case #261) sees a "changed" hash and cannot tell whether the config is corrupt or merely different.
- **The silent fallback defeats the endpoint's audit purpose**: the doc says the fingerprint is for "release review and upgrade planning" and "incident response"; a silent 0 is exactly the wrong signal in an incident.
- **The error path is untested**: no test covers `compute_config_version_hash` failing inside the fingerprint, so the fallback behavior is undefined-but-shipped.

## Root cause

The `unwrap_or(0)` was added as a defensive fallback (the comment says "If config is unreadable, fall back to sentinel 0") without a way for consumers to distinguish the sentinel from real data.

1. The fix options: (a) propagate the error (`?`) — changes the endpoint's contract (it currently documents `NotInitialized` as its only error and "All other contract states return a valid fingerprint"); (b) add a distinct sentinel (e.g. `u64::MAX`) plus a `config_readable: bool` field — an additive ABI change to `ContractStateFingerprint` (stability-guarded).
2. The fingerprint is designed to be callable in pre-migration states (`needs_migration: true`), where config reads may legitimately be partial; the fix must define what "unreadable" means per state.
3. `compute_config_version_hash` itself can fail per-severity (`load_config` → `ConfigNotFound`), so the fingerprint must decide the policy for a partially-missing config.

## Acceptance criteria

- [ ] A config-hash failure inside the fingerprint is distinguishable from a valid hash.
- [ ] The endpoint's documented error contract matches the chosen behavior.
- [ ] Tests cover the unreadable-config path.

## Out of scope

The hash-collision analysis (companion issue) and the custom-severity exclusion (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`get_contract_state_fingerprint`, `compute_config_version_hash`).

Mirabel64:
Fixes #491
Fixes #492
Fixes #493
Fixes #494

One PR closing all four issues assigned to Mirabel64.

## #491 — scratch_test.rs cleanup + orphan-module lint

- Removed the leftover `apexchainx_calculator/scratch_test.rs` (never declared as a module, never compiled, its `test_sym` helper a trap for contributors).
- Added `scripts/check-orphan-modules.sh`: fails when any `.rs` file under the crate is never declared as a module — a file invisible to `cargo check`/`clippy`/`test` rots silently. Wired into CI as an early gate in `client-checks` and exposed as `just lint-orphans` (added to `just ci`).
- Contributor docs now state where throwaway code belongs: `CODING_STYLE.md` Part 4 (module-declaration + scratch-code policy) and a new bullet in `CONTRIBUTING.md` Step 2.
- The src/ orphans tracked by companion issue #422 (restored in PR #545) are allowlisted with a comment; the lint warns when the allowlist goes stale, so it self-cleans once #545 lands.

## #492 — get_public_api descriptor completeness guardrail

- Added `CANONICAL_PUBLIC_METHODS` (the impl-truth list, 61 methods with documented mutates/auth/event metadata) plus two tests in `tests.rs`:
  - `test_492_descriptor_covers_every_public_method` — fails when a public `#[contractimpl]` method is absent from `get_public_api`.
  - `test_492_descriptor_lists_only_declared_methods` — fails when the descriptor lists a method the impl no longer exposes.
- The section comment documents the three-way sync rule (impl block ⇄ `get_public_api` ⇄ canonical list) as the enforcement mechanism.

## #493 — pause_info Option workaround standardization

- Documented the single convention governing "absent" state across `#[contracttype]` surfaces in `CODING_STYLE.md` Part 3: `Vec<T>` with a max-length invariant (empty = absent, one element = present); getters keep `Result<Option<T>, _>`. Referenced from `CONTRIBUTING.md`.
- Enforced the `AuditState.pause_info` invariant with `test_audit_state_pause_info_invariant_across_pause_cycle` (empty when unpaused, exactly one when paused, always matching `get_pause_info`), and updated the field docs to carry the `INVARIANT: len() <= 1` note.

## #494 — fingerprint no longer masks a corrupt config as hash 0

- `get_contract_state_fingerprint` now propagates a config-hash failure (`?` instead of `unwrap_or(0)`): a legitimate `config_version_hash` is never 0 (polynomial seed is 1), so the old sentinel made a corrupt config indistinguishable from a valid state.
- Error contract updated in the doc comment (`NotInitialized`/`ConfigNotFound` when the config is unreadable; all *readable* states still return a valid fingerprint, including pre-migration).
- New tests: `test_494_fingerprint_fails_when_config_key_missing`, `test_494_fingerprint_fails_when_config_partially_missing`, and `test_494_fingerprint_hash_never_zero_on_readable_config`.

## Verification

- `cargo test --lib` — 598 passed, 0 failed (9 new/related tests pass)
- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo fmt --check` — clean
- `cargo check --target wasm32-unknown-unknown --lib` — clean (no_std)
- `./scripts/check-no-std.sh` and `./scripts/check-orphan-modules.sh` — pass (lint verified to fail on a new orphan)
- TS parity fixtures unchanged (`ts/` has no diff)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,14 @@ jobs:
       - name: no_std compatibility lint (early gate)
         run: ./scripts/check-no-std.sh
 
+      # Issue #491: Orphan module lint — fails when a `.rs` file in the crate
+      # is never declared as a module (e.g. the committed scratch_test.rs that
+      # shipped for weeks without CI noticing). An undeclared file is invisible
+      # to cargo check/clippy/test, so its tests rot silently; this gate closes
+      # that class.
+      - name: Orphan module lint (early gate)
+        run: ./scripts/check-orphan-modules.sh
+
       # Issue #12: WASM no-std compliance check.
       # Compiles only the library crate for wasm32-unknown-unknown so that any
       # accidental `std` import surfaces here (before deployment) as a compile
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -106,6 +106,12 @@
   - Performs configuration parameter validation (SC-W5-046)
 - `get_stats` now returns a `SLAStats` struct; callers should use field access rather than tuple destructuring
 - History entries returned by `get_history` include `schema_version` for result envelope versioning
+- `get_contract_state_fingerprint` no longer reports `config_version_hash = 0` when the config is unreadable. Previously an unreadable `CONFIG_KEY` (storage corruption, bad migration) was masked with `unwrap_or(0)`, making a corrupt config indistinguishable from a legitimate hash — which is never 0 — and defeating the endpoint's incident-response purpose. A config-hash failure now propagates as `NotInitialized`/`ConfigNotFound`, and the documented error contract was updated to match; tests cover the missing-key and partially-missing-config paths (#494)
+
+### Added
+- `scripts/check-orphan-modules.sh` — orphan-module lint that fails when a `.rs` file under the crate is never declared as a module (an undeclared file is invisible to cargo check/clippy/test, so its tests rot silently). Wired into CI as an early gate and available as `just lint-orphans`. The leftover `apexchainx_calculator/scratch_test.rs` was removed as part of this (#491)
+- `get_public_api` descriptor completeness guardrail — `test_492_descriptor_covers_every_public_method` and `test_492_descriptor_lists_only_declared_methods` compare the hand-enumerated descriptor against the canonical `CANONICAL_PUBLIC_METHODS` list in both directions, so a public method missing from the descriptor or a descriptor entry that no longer exists on the impl fails CI (#492)
+- `AuditState.pause_info` invariant enforcement — a test pins the `Vec`-stands-in-for-`Option` convention (empty when unpaused, exactly one element when paused, always matching `get_pause_info`), and `CODING_STYLE.md` Part 3 documents the single convention governing absent state across `#[contracttype]` surfaces; Part 4 documents the module-declaration and scratch-code policy (#493)
 
 ---
 
--- a/CODING_STYLE.md
+++ b/CODING_STYLE.md
@@ -282,4 +282,107 @@ If a property is expressed in a test and in a comment, the comment is likely to 
 
 This policy is enforced through the PR review checklist above. Reviewers must verify compliance with each category before approving. Automated enforcement via `cargo clippy` lints (`missing_docs`, `clippy::missing_errors_doc`) is enabled for public items; violations block CI.
 
+---
+
+## Part 3 — Optional-State Convention for `#[contracttype]` Surfaces (Issue #493)
+
+This section defines the single convention for representing **absent** state
+across `#[contracttype]` struct fields, so consumers never have to learn two
+shapes for the same semantic.
+
+### The SDK limitation
+
+`#[contracttype]` structs cannot contain `Option<T>` fields: the SDK's ScVal
+conversion for a derived `#[contracttype]` struct has no `From<&Option<T>>`
+impl, so a struct with an `Option` field does not compile. This is a platform
+limitation, not a design choice — every surface hits it independently and
+must apply the same rule.
+
+### The convention
+
+1. **`#[contracttype]` struct fields model absent state with `Vec<T>`.**
+   Empty = absent; single-element = present. Every such field must carry a
+   comment stating its **max-length invariant** (e.g. `len() <= 1`) so the
+   "one element max" rule is reviewable, plus the invariant must be pinned by
+   a test.
+2. **Getters may return `Result<Option<T>, _>`.** Function return values are
+   not subject to the `#[contracttype]` field limitation, so the read surface
+   (`get_pause_info`, `get_pending_admin`, `get_pending_operator`) uses
+   `Option` directly. The `Vec` field in an aggregate struct is the *wire*
+   representation; the `Option` getter is the *semantic* one, and they must
+   agree (empty ⇔ `None`, one element ⇔ `Some(_)`).
+3. **Never invent a second representation.** If a field can be absent, do not
+   use a sentinel value, a zero-address, or a `bool` flag alongside a `Vec` —
+   use the `Vec` convention and document the invariant.
+
+### Example
+
+```rust
+/// Pause metadata when paused, empty otherwise. `Vec` stands in for `Option`
+/// because `#[contracttype]` cannot convert `Option<PauseInfo>` (#493).
+/// INVARIANT: `pause_info.len() <= 1` — empty = unpaused, one element = paused.
+pub pause_info: Vec<PauseInfo>,
+```
+
+The invariant above is enforced by `audit_state::tests::
+test_audit_state_pause_info_invariant_across_pause_cycle`, which asserts
+empty ⇔ unpaused and exactly-one ⇔ paused, and that both match
+`get_pause_info()`.
+
+### Review checklist
+
+- [ ] Every new `#[contracttype]` struct field that can be absent uses `Vec<T>`
+      with a documented max-length invariant (Part 3 convention), never a
+      sentinel or a parallel `bool`.
+- [ ] The aggregate field and its `Option`-typed getter agree, pinned by a test.
+- [ ] Changing a `#[contracttype]` field's shape (e.g. `Vec<PauseInfo>` → a
+      wrapper) is an ABI change and must go through the stability process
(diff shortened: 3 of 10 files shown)


usmanimamu17-create (approved):
Solid work, merging.

PR #552, opened 2026-08-31
payload optimization

### LINKED ISSUE #434: The SC-W5-037 payload optimization is documented but never applied: events still emit payment_type while the optimizer says it is derivable
## Problem

`src/payload_optimizer.rs` (orphan module — see companion issue) documents SC-W5-037:

```rust
//! 1. **Derive `payment_type` from `status`** — "viol" → "pen", "met" → "rew"
//!    eliminates the need to store payment_type separately in events
```

But the live publish paths in `src/lib.rs` emit `payment_type` in both `sla_calc` and `set_int` payloads, and the `SLAResult` struct stores it. The optimizer's "Backend Guidance" tells consumers they can derive `payment_type` from `status` — which is true of the current data (the invariant `derive_payment_type(status) == payment_type` holds) — but the optimization itself (dropping the field) was never applied.

Consequences:

- **The optimization's cost reduction never happened**: the module's stated purpose is "reduces on-chain storage and gas costs", but no event or struct was changed, so the documented benefit is unrealized while the helper code ships.
- **The guidance and the ABI are inconsistent in contract**: backends told they can "eliminate the need to store both fields" will still receive both fields in every event, and the optimizer's `is_consistent_payment` check exists to validate an invariant nothing validates at the publish site.
- **The invariant is unenforced**: nothing asserts `derive_payment_type(status) == payment_type` when emitting, so if a future change emits an inconsistent pair, consumers following the derivation guidance will silently misread amounts.

## Root cause

The optimizer module was written as a spec/helper and merged; the publish paths and `SLAResult` were never refactored to adopt it.

1. Actually dropping `payment_type` from the events is a breaking ABI change (field removal per `docs/EVENT_COMPATIBILITY_POLICY.md` requires a version bump to v2) and requires every indexer to switch to derivation — coordinated across the ecosystem.
2. The conservative fix is to enforce the invariant at publish time (assert/emit-error on inconsistency) and document that the derivation guidance holds, deferring the field removal.
3. The relationship between `SLAResult` (which stores `payment_type` for history queries) and events (which could derive it) means the struct and the events would diverge — history read paths still need the stored field even if events drop it.

## Acceptance criteria

- [ ] Either the events adopt the optimization (with a version bump and migration note) or the optimizer module's claims are corrected to match reality.
- [ ] The invariant `derive_payment_type(status) == payment_type` is enforced at the publish path (or the guidance is removed).
- [ ] Consumers following the documented guidance cannot be misled about the payload contents.

## Out of scope

The orphan-module cleanup (companion issue) and reducing the three-event payload redundancy (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/payload_optimizer.rs`, `apexchainx_calculator/src/lib.rs` (`publish_sla_event`, `publish_settlement_intent_event`).

### LINKED ISSUE #435: EventPublisher is dead code: the live publish paths implement the topic layout outside the documented canonical publisher
## Problem

`src/event.rs` defines `EventPublisher` as the canonical, stateless publisher:

```rust
//! Events are always published through this struct to ensure consistent
//! topic layout and payload formatting.
pub struct EventPublisher;
impl EventPublisher {
    pub fn publish_calculation_executed(env: &Env, topic: Symbol, ...) { ... }
}
```

But nothing calls it: the real events are published inline in `src/lib.rs` (`publish_sla_event`, `publish_settlement_intent_event`, `publish_duplicate_input_event`, and the direct `env.events().publish(...)` calls in `set_config`, `pause`, `migrate`, etc.). `publish_calculation_executed` is not even the shape the contract emits (`sla_calc` uses a 7-field tuple; `publish_calculation_executed` publishes a `CalculationExecutedEventV1` struct payload).

Consequences:

- **The "always published through this struct" guarantee is false**: the contract's events flow through at least four different inline publish sites with hand-written tuples; the canonical publisher is never exercised, so its "consistent topic layout" promise is untested against reality.
- **`CalculationExecutedEventV1` documents a payload that does not exist on-chain**: the struct defines a field order (`input_key`, `input_value`, `result_value`, `timestamp`) that matches no emitted event — a consumer or reviewer who trusts `event.rs` as the schema source will decode the real events incorrectly.
- **The doc-comment contract in `calculation.rs` points at a fiction**: it says "See `crate::event::EventPublisher` for the event publication abstraction", but the abstraction is unused.

## Root cause

`EventPublisher` was authored as the intended publisher before the inline publish paths were written; the paths were merged without adopting it, and `event.rs` was never deleted or wired.

1. Two resolutions: route all publishes through `EventPublisher` (a refactor of the publish helpers — behavior-neutral if the tuples match, which `publish_calculation_executed` currently does not) or delete `event.rs` and move the schema documentation into `event_schema.rs` where the real shapes live.
2. `CalculationExecutedEventV1` is `#[contracttype]` — part of the (unguarded) ABI surface; removing it or changing it must follow the stability process.
3. The publisher refactor touches every event-emitting path and their tests, so the change is a good vehicle for the shared field-order canonicalization (companion issue) but must stay behavior-neutral to be reviewable.

## Acceptance criteria

- [ ] Either all contract events are published through `EventPublisher` (with payload shapes matching the real ABI), or `event.rs` is removed and its schema content relocated to `event_schema.rs`.
- [ ] No doc-comment points to `EventPublisher` as the canonical publisher while the live paths bypass it.
- [ ] Event-shape tests pass unchanged (behavior-neutral).

## Out of scope

The payload field-order divergence across events (companion issue) and per-event size budgets.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/event.rs`, `apexchainx_calculator/src/lib.rs` (publish helpers), `apexchainx_calculator/src/calculation.rs` (doc reference).

### LINKED ISSUE #436: History trimming drops the oldest entry with no event: indexers reconstructing history from events silently lose the evicted result
## Problem

When `calculate_sla` appends beyond the retention limit, the oldest entry is silently dropped:

```rust
// src/lib.rs (and src/calculation.rs, identical)
if history.len() > retention_limit {
    let mut trimmed = Vec::new(env);
    for i in 1..history.len() { trimmed.push_back(history.get(i).unwrap()); }
    env.storage().instance().set(&HISTORY_KEY, &trimmed);
}
```

The evicted `SLAResult` is discarded without any event. The `pruned`/`pruned_a` events only fire for admin-initiated prunes.

Consequences:

- **Event-based history reconstruction diverges from on-chain state**: a backend that builds its SLA history by replaying `sla_calc` events (the documented indexing path) will retain entries the contract has evicted; after enough traffic it holds a longer history than the contract, and its aggregate stats disagree with `get_stats`.
- **The eviction is invisible to reconciliation**: nothing tells indexers or auditors which decision was dropped, so the "retention window" is an implementation detail rather than a documented, observable boundary.
- **The 1000-entry cap's effect on the event stream is undefined**: consumers cannot tell whether their event backlog is complete relative to the contract's retained window.

## Root cause

Trimming was implemented as a storage optimization; the event stream — the contract's documented integration surface for indexers — was not extended to announce evictions.

1. Emitting an eviction event means defining its schema (e.g. a `hist_trim` event carrying the dropped entry or its `outage_id`/`recorded_at`): a new event name is an event-ABI addition that must follow the SC-099 checklist and update the catalogs (companion issues).
2. Alternatively, the retention behavior can be redefined so evictions happen only via explicit `prune_history` (which already emits events), making the window observably bounded — but that changes storage-growth behavior and interacts with `MAX_RECALCS_PER_OUTAGE`.
3. The payload decision (full dropped entry vs. reference) trades auditability against event size; `offchain/eventSizeRegression.ts` shows the project already budgets event sizes, so the payload must respect those budgets or the budgets must change.

## Acceptance criteria

- [ ] Every eviction from the retained window is observable in the event stream (or evictions only occur through event-emitting prune paths).
- [ ] Replaying the event stream reconstructs the contract's retained history exactly.
- [ ] The eviction payload and event catalog are updated per the event-change checklist.

## Out of scope

Redesigning history storage (companion issue) and the admin-prune event payload inconsistency (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`calculate_sla` trimming block), `apexchainx_calculator/src/history.rs` (`prune_history`, `prune_history_by_age`).

### LINKED ISSUE #437: initialize emits no event: contract bootstrap (roles and default config) is invisible to indexers
## Problem

`initialize` (`src/lib.rs`) writes roles, stats, telemetry keys, and the default config — and emits nothing:

```rust
    if env.storage().instance().has(&ADMIN_KEY) { return Err(SLAError::AlreadyInitialized); }
    admin.require_auth(); operator.require_auth();
    ... // CONFIG_KEY, HISTORY_KEY, STATS_KEY, telemetry keys
    Self::write_version(&env);
    Ok(())
}
```

The contract's first event is the first `sla_calc`/`cfg_upd`/etc. from a later call.

Consequences:

- **Indexers have no bootstrap anchor**: consumers that build state from the event stream (the documented integration path) cannot learn the initial admin, operator, or default config from events; they must call read endpoints, and they have no way to detect that a deployment happened at all.
- **Replay-based tooling cannot reconstruct the initial state**: the audit trail for "who initialized the contract and with which defaults" exists only as on-chain storage, not as an event.
- **Deployment monitoring is blind**: a contract created and never used is indistinguishable from one that was never created, because `healthcheck` only reports version/init status and no event marks creation.

## Root cause

`initialize` predates the event-emission conventions; the catalog (`src/event_schema.rs`) has no "initialized" event, and the SC-100 checklist was applied to later methods only.

1. Adding an `init_done`-style event is an event-ABI addition (new name + payload, e.g. admin/operator addresses) that must follow the SC-099 checklist and the catalogs in `event_schema.rs`/`api_stability.rs`/lib.rs — the same single-source-of-truth problem as the catalog companion issues.
2. The payload carries `Address` values (not Symbols), which the current event family only uses in topic[2]; adding addresses to a payload is a new shape and must be documented for indexers.
3. `get_public_api` lists `initialize` with an empty event column; the descriptor must be updated in lockstep so discovery consumers learn about the new event.

## Acceptance criteria

- [ ] `initialize` emits a versioned bootstrap event carrying the initial roles (and optionally the config hash).
- [ ] The event appears in all three event catalogs and the API descriptor.
- [ ] Replaying events from a fresh deployment reconstructs the initial admin/operator.

## Out of scope

Emission of an initial config snapshot event and event-driven config reconstruction (companion issues).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`initialize`, `get_public_api`), `apexchainx_calculator/src/event_schema.rs`.

ahmadogo:
## Description

## Type of Change
- [ ] Documentation update

## Related Issues

## Changes Made
- Closes #434
- Closes #435 
- Closes #436 
- Closes #437 
- 
- 

## Testing

## Checklist


usmanimamu17-create (approved):
Nice optimization, merging 👍

PR #553, opened 2026-08-31
Feat/severity

### LINKED ISSUE #430: Event-name constants exist in three files with no single source of truth: the lib.rs catalog omits cfg_frz, cfg_unfrz, and migrate_done
## Problem

Event names are declared or enumerated in three places that do not agree:

- `src/lib.rs` — the `EVENT_*` constants used by the contract, plus a "===== Event Payload Schemas =====" comment catalog.
- `src/event_schema.rs` — the documented catalog (21 constants including `EVENT_VERSION`).
- `src/api_stability.rs` — `event_name_symbols()` (16 names, see companion issue).

The lib.rs comment catalog lists `sla_calc`, `cfg_upd`, `paused`, `unpause`, `op_set`, `pruned`, `pruned_a`, the six governance events, `set_int`, `dup_input`, and `stats_sat` — but omits `cfg_frz`, `cfg_unfrz`, and `migrate_done`, even though lib.rs itself defines the freeze constants and publishes the migrate event.

Consequences:

- **Adding an event requires coordinated edits in three files**: the SC-099 checklist demands versioning discipline, but the mechanical burden (constant + catalog comment + schema doc + stability list) is spread across files with no shared source, so one is always stale — here, the lib.rs catalog already is.
- **The "Full schema documentation" pointer is incomplete**: lib.rs's comment says "Full schema documentation: event_schema.rs", yet the inline catalog — the first thing a contributor reads — is missing three events the code emits.
- **The freeze events have no doc coverage at all**: `cfg_frz`/`cfg_unfrz` are emitted by `freeze_config`/`unfreeze_config` and documented in `event_schema.rs`, but the reviewer-facing catalog in lib.rs omits them, so a review of "all events this contract emits" from lib.rs is wrong.

## Root cause

The comment catalog predates the freeze and migration events; it was never re-synced, and nothing enforces parity between the comment, the constants, and the schema module.

1. The durable fix is one canonical catalog (e.g. a `const` array in `event_schema.rs` listing name/version/payload-shape) with the other two derived or asserted against it — the same single-source-of-truth problem as the stability guardrail (companion issue).
2. The lib.rs comment is documentation, not code; making it verifiable means either converting it to a test that checks each `EVENT_*` constant is documented, or removing the inline catalog in favor of the schema module.
3. Any change must not reorder or re-version existing events while the catalog is consolidated.

## Acceptance criteria

- [ ] One file is the canonical event catalog; the lib.rs comment either derives from it or is removed.
- [ ] A test fails when a new `EVENT_*` constant is added without a catalog entry.
- [ ] All 20 emitted event names appear in the canonical catalog.

## Out of scope

The 16-vs-20 stability list gap (companion issue) and `EVENT_MIGRATE_DONE`'s `&str`-vs-`Symbol` type.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (event constants + catalog comment), `apexchainx_calculator/src/event_schema.rs`.

### LINKED ISSUE #431: pause overwrites existing PauseInfo when already paused: the previous reason and timestamp are silently lost
## Problem

`pause` in `src/metadata.rs` (and the identical inline version in `src/lib.rs`) is not idempotent:

```rust
env.storage().instance().set(&PAUSED_KEY, &true);
env.storage().instance().set(&PAUSE_INFO_KEY, &PauseInfo { reason, paused_at, paused_by: caller.clone() });
env.events().publish((EVENT_PAUSED, EVENT_VERSION, caller.clone()), (true,));
```

Calling `pause` while already paused overwrites `PAUSE_INFO_KEY` and emits another `paused` event; the original reason/timestamp/caller are gone.

Consequences:

- **The pause history is destroyed**: `get_pause_info` is the only window into why the contract is paused; a second `pause` (e.g. an incident response that pauses again with a new reason) erases the first reason and timestamp, so auditors see only the latest reason.
- **Backends cannot reconstruct the pause timeline**: the `paused` event carries no timestamp or reason payload (only `(true,)`), so event consumers cannot recover what `get_pause_info` lost.
- **A no-op transition is indistinguishable from a real one**: re-pausing emits the same event as the initial pause, so indexers counting pause events will over-count.

## Root cause

The pause path was written as a simple state set without checking the current state; the `PauseInfo` overwrite is a side effect of that.

1. Two coherent designs: (a) make `pause` idempotent — return an error (e.g. a new `AlreadyPaused` variant) or `Ok` without touching `PAUSE_INFO_KEY` when already paused; (b) keep overwrite semantics but emit a distinct event and preserve history in the payload. The choice changes the error surface and event ABI.
2. Adding an `AlreadyPaused` error variant is an `SLAError` ABI change (the enum is stability-guarded; the failure catalog in `get_failure_schema` and `error_responses.rs` must be updated — see the error-catalog companion issue).
3. The `unpause`-when-unpaused path has the same asymmetry, so the fix should define the full pause state machine (idempotency, events, errors) in one place.

## Acceptance criteria

- [ ] Re-pausing does not silently erase the prior `PauseInfo`.
- [ ] The event stream and `get_pause_info` agree on the pause timeline.
- [ ] The pause/unpause state machine (idempotent calls, no-op transitions) is documented and tested.

## Out of scope

A full pause-history store and the `unpause` no-op event (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/metadata.rs`, `apexchainx_calculator/src/lib.rs` (`pause`), `apexchainx_calculator/src/event_schema.rs` (paused/unpause payloads).

### LINKED ISSUE #432: unpause and pause emit events even when no state transition occurs: backends cannot filter no-op lifecycle events
## Problem

Both lifecycle methods emit unconditionally:

```rust
// src/metadata.rs
pub fn unpause(env: &Env, caller: &Address) -> Result<(), SLAError> {
    ...
    env.storage().instance().set(&PAUSED_KEY, &false);
    env.storage().instance().remove(&PAUSE_INFO_KEY);
    env.events().publish((EVENT_UNPAUSED, EVENT_VERSION, caller.clone()), (false,));
    Ok(())
}
```

Calling `unpause` on an already-unpaused contract (or `pause` on an already-paused one) writes identical state and emits an event. `PAUSED_KEY`'s previous value is never read.

Consequences:

- **Event-based state machines diverge from storage**: an indexer that replays events to reconstruct pause state will toggle on every call, while the contract's storage toggles only on real transitions; a redundant `unpause` call flips the indexer's state to unpaused regardless of the contract's actual state.
- **Operational alerts fire on no-ops**: tooling that alerts on `paused` events will page operators for a pause that changed nothing.
- **The event stream overstates activity**: governance and audit trails cannot distinguish an actual transition from a redundant call, since both emit the identical payload.

## Root cause

The methods write-then-emit without reading the prior `PAUSED_KEY` value; no transition guard exists.

1. Making emissions conditional on the previous value is a behavior change for consumers that currently tolerate the redundancy; it must be decided whether a no-op call returns an error, returns `Ok` silently, or emits a distinct "no-op" signal.
2. The same transition-detection question applies to `freeze_config`/`unfreeze_config` and `set_config` (no-op writes bump the config hash — companion issue), so a consistent "transitional event" policy should cover all lifecycle methods.
3. The `paused`/`unpause` payloads are fixed `(true,)`/`(false,)`; if a no-op distinction is added, it must be additive or version-bumped per the event policy.

## Acceptance criteria

- [ ] A no-op `pause`/`unpause` call does not emit a transition event (or emits a clearly distinct one).
- [ ] Replaying the event stream reconstructs `get_pause_info` state exactly.
- [ ] The chosen no-op semantics are documented for consumers.

## Out of scope

The `PauseInfo` overwrite on re-pause (companion issue) and an `AlreadyPaused` error variant.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/metadata.rs`, `apexchainx_calculator/src/lib.rs` (`pause`, `unpause`).

### LINKED ISSUE #433: deprecated_symbols and severity_aliases are always empty in get_result_schema: the deprecation machinery consumers are told to rely on is never populated
## Problem

`get_result_schema` (`src/lib.rs`) always returns empty deprecation data:

```rust
deprecated_symbols: Vec::new(&env),
severity_aliases: Vec::new(&env),
```

Meanwhile `src/event_schema.rs` documents a "Symbol Deprecation Protocol" telling backends to "MUST check `deprecated_symbols` at startup and log warnings for any deprecated symbols they still rely on", and `src/lib.rs`'s `SLAResultSchema` doc says `severity_aliases` lets backends "translate old severity symbols to current ones" (#239).

Consequences:

- **The documented startup check is a no-op**: backends that implement the recommended `deprecated_symbols` scan receive an empty list forever; when a symbol is actually deprecated later, the schema endpoint will not signal it unless someone remembers to populate the list — the protocol has no teeth.
- **Historical interpretation is unsupported**: `SeverityAliasMapping` exists to translate old severity symbols in historical `SLAResult` entries, but no code path ever emits a mapping; a backend following the docs has no way to interpret pre-rename history.
- **The two mechanisms are pure scaffolding**: `DeprecatedSymbol` and `SeverityAliasMapping` structs, their schema fields, and the docs exist, but the contract never produces a single entry, so the features cannot be exercised end-to-end.

## Root cause

The deprecation/alias machinery was built as API surface ahead of any actual symbol change; nothing populates it, and no test asserts it is ever populated (so the empty default is unverified behavior).

1. Populating `deprecated_symbols` correctly requires a deprecation registry the contract maintains across releases — the "deprecated_at"/"removal_version" lifecycle implies the schema is static per binary but historical events keep old symbols, which means the data belongs in docs/off-chain tooling as much as in the endpoint.
2. `severity_aliases` only matters when a severity is renamed; the contract has no rename facility today, so the field can only be populated by (a) a future rename feature or (b) a documented mapping shipped in a release — both are design decisions.
3. The fix must decide: either wire real data (deprecate something, or add a `set_deprecated_symbols`-style admin path — a storage and governance addition), or downgrade the docs so consumers do not rely on the fields.

## Acceptance criteria

- [ ] The schema endpoint either returns populated deprecation/alias data, or the docs are corrected so consumers do not depend on it.
- [ ] A test covers the population path (real or admin-driven) end-to-end.
- [ ] `docs/SEVERITY_ALIAS_POLICY.md` and the event deprecation protocol match the implemented behavior.

## Out of scope

Introducing an actual severity rename feature.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`get_result_schema`, `SeverityAliasMapping`), `apexchainx_calculator/src/event_schema.rs` (deprecation protocol).

ayshadogo:
## Description

## Type of Change
- [ ] Documentation update

## Related Issues

## Changes Made
- Closes #430 
- Closes #431 
- Closes #432 
- Closes #433 
- 
- 

## Testing

## Checklist


usmanimamu17-create (approved):
LGTM 🚀

PR #555, opened 2026-08-31
DeploymentPolicy implemented

### LINKED ISSUE #442: rollback_all ignores compensation failures: a failed compensation leaves the workflow partially rolled back with no signal
## Problem

`CrossContractSafety::rollback_all` (`src/cross_contract_safety.rs`) invokes every compensation and discards the results:

```rust
pub fn rollback_all(&mut self, env: &Env) {
    while let Some((contract_id, _function_name, action)) = self.compensation_stack.pop_back() {
        let _ = env.try_invoke_contract::<Val, Val>(&contract_id, &action.tag, action.args.clone());
    }
}
```

Consequences:

- **A failed compensation is invisible**: if the compensation call itself errors (target paused, version mismatch, args invalid), the workflow is left in a partially-compensated state and the caller — who already returned `Err` from the original failure — has no way to know rollback was incomplete.
- **The "failure rollback semantics" guarantee is overstated**: the module doc promises "any partial state changes must be rolled back"; the implementation rolls back *best-effort*.
- **The compensation stack is consumed regardless**: `pop_back` empties the stack even when compensations fail, so retrying rollback is impossible — the recovery state is destroyed by the failed attempt.

## Root cause

Rollback was written as a fire-and-forget unwind loop; the doc's strong guarantee was not matched by error handling.

1. There is no storage to persist "compensation pending" state in this model — the stack is in-memory per call — so surfacing failures means returning a structured result from `rollback_all` (e.g. which compensations failed) and deciding whether the caller can retry.
2. In a Soroban transaction, a failed cross-contract call aborts the whole transaction anyway; the model's real value is ordering and determinism, so the design must clarify what "rollback" means when the host already guarantees atomicity — the failure surface may be narrower than the docs imply.
3. The `Compensated` status variant exists but is never produced by any code path (only asserted in tests), suggesting the intended "compensation applied" reporting was never built.

## Acceptance criteria

- [ ] `rollback_all` reports which compensations failed (or the model documents that rollback is best-effort within a transaction).
- [ ] The `Compensated` status is either produced by the code or removed as dead vocabulary.
- [ ] Tests cover a failing compensation and assert the reported outcome.

## Out of scope

Persisting rollback state across transactions and the real cross-contract wiring (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/cross_contract_safety.rs` (`rollback_all`, `CrossContractCallStatus`).

### LINKED ISSUE #443: CrossContractSafety.call registers compensation for RecoverableError while requires_rollback says recoverable errors need no rollback: the two helpers disagree
## Problem

The two public helpers in `src/cross_contract_safety.rs` encode opposite semantics for recoverable errors:

```rust
// call(): a recoverable error is treated as a successful step — compensation is registered
CrossContractCallStatus::Success | CrossContractCallStatus::RecoverableError => {
    self.compensation_stack.push_back((...));
    Ok(result)
}

// requires_rollback(): recoverable errors do NOT require rollback
pub fn requires_rollback(status: CrossContractCallStatus) -> bool {
    status == CrossContractCallStatus::FatalError || status == CrossContractCallStatus::DispatchFailed
}
```

Consequences:

- **The model has no coherent recoverable-error policy**: `call` keeps going (and registers compensation) after a `RecoverableError`, while `requires_rollback` says such errors leave the workflow safe to continue without compensation — a caller cannot tell whether a recoverable step was "done" or "not done" from the returned status.
- **The compensation stack can grow from failed calls**: if a recoverable error means the step did not happen, registering a compensation for it is wrong (rollback would reverse a step that never executed); if it means the step happened with degraded semantics, `requires_rollback` is wrong.
- **A subsequent fatal error triggers rollback of the recoverable step too**: `rollback_all` unwinds the whole stack including entries pushed for recoverable errors, potentially double-reversing.

## Root cause

The two functions were written against different mental models (status classification vs. compensation bookkeeping) and never reconciled; the tests assert each in isolation, so the contradiction ships.

1. Resolving it requires defining what `RecoverableError` means for the workflow state: did the call commit or not? The safe answer depends on the target contract's semantics, which the generic `safe_invoke_contract` cannot know — so the model may need a per-call "committed on recoverable" flag.
2. The `call` signature and `SafeCallResult` are public library API; changing the status/compensation mapping changes the coordination harness expectations.
3. The contradiction is exactly the kind of semantic drift the (unwired) module needs before real cross-contract calls land, so the fix should be paired with an integration-style test using a mock peer.

## Acceptance criteria

- [ ] `call` and `requires_rollback` agree on the recoverable-error policy, documented in the module doc.
- [ ] Compensation is registered only for steps that actually committed.
- [ ] Tests pin the behavior for all five `CrossContractCallStatus` variants.

## Out of scope

The escrow/settlement peers and the error-payload preservation (companion issues).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/cross_contract_safety.rs` (`call`, `requires_rollback`).

### LINKED ISSUE #444: FN_LOCK_FUNDS, FN_RELEASE_PAYMENT, and the COMP_* tags describe an escrow interface with no consumer in this repo
## Problem

`src/cross_contract_safety.rs` defines the escrow interface the safety model assumes:

```rust
pub const FN_LOCK_FUNDS: Symbol = symbol_short!("lock_fnds");
pub const FN_RELEASE_PAYMENT: Symbol = symbol_short!("rel_pay");
pub const FN_CANCEL_SETTLEMENT: Symbol = symbol_short!("can_setl");
pub const COMP_UNLOCK_FUNDS: Symbol = symbol_short!("unlck_fnd");
pub const COMP_REVERSE_SETTLE: Symbol = symbol_short!("rev_setle");
```

No contract method in this repo calls `lock_fnds`/`rel_pay`/`can_setl`, and no test registers a peer implementing them. The `contract_name` constants in `src/version_negotiation.rs` (`pay_escro`, `settle`) point at contracts that exist only as names.

Consequences:

- **The compensation tags are unverifiable**: `unlck_fnd`/`rev_setle`/`unp_escro` are the actions `rollback_all` would invoke, but no implementation or mock validates that these function names match the real escrow contract's ABI — a single rename in the escrow repo silently breaks rollback.
- **The interface is a frozen assumption**: if the ecosystem's escrow contract exposes different function names (or requires different arguments), the mismatch is discovered only at integration time, not by CI.
- **`get_contract_info` advertises `safe_call` while the interface it protects is fictional** (see the feature-list companion issue).

## Root cause

The escrow interface constants were authored alongside the safety model as placeholder names for a planned peer contract; no cross-repo check exists.

1. Verifying the interface requires either the real escrow contract (cross-repo dependency — likely in the ApexChainx ecosystem's other repos) or a registered mock contract in tests that implements `lock_fnds`/`rel_pay` and asserts the safety model calls them with the right args.
2. The function names are 9-char `symbol_short` truncations of English words — a convention the peer must match exactly; documenting the ABI contract (names + arg types) is the deliverable that makes the peer implementable.
3. If the escrow contract does not exist yet, the issue is to document the interface as provisional and add a compile-time/test-time registry that fails when the constants are used without a matching mock.

## Acceptance criteria

- [ ] A test registers a mock peer implementing `lock_fnds`/`rel_pay`/`can_setl` and exercises the compensation flow against it.
- [ ] The expected function-name and arg contract is documented for the escrow implementer.
- [ ] The `safe_call` feature flag and interface constants are reconciled with what actually exists.

## Out of scope

Building the escrow/settlement contracts and wiring `calculate_sla` to them (companion issues).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/cross_contract_safety.rs` (interface constants), `apexchainx_calculator/src/version_negotiation.rs` (contract names).

### LINKED ISSUE #445: DeploymentPolicy::verify_deployment_compatibility hardcodes >= 1 instead of using REQUIRED_PROTOCOL_VERSION and always returns true on any live ledger
## Problem

`src/deployment_policy.rs` (orphan module — see companion issue) declares the constant but checks a literal:

```rust
pub const REQUIRED_PROTOCOL_VERSION: u32 = 1;

pub fn verify_deployment_compatibility(env: &Env) -> bool {
    let current_protocol = env.ledger().protocol_version();
    current_protocol >= 1   // <- hardcoded, ignores REQUIRED_PROTOCOL_VERSION
}
```

Consequences:

- **The constant is decorative**: changing `REQUIRED_PROTOCOL_VERSION` to 2 would have no effect on the check; the guard is a tautology.
- **The check is meaningless on every live ledger**: Soroban ledgers run protocol versions far above 1 (protocol 20+ is current), so `>= 1` is always `true`; the "deployment compatibility verification" gives operators no protection and no signal.
- **The module's promise ("verifies that the target ledger meets minimum version requirements") is unfulfilled**: any protocol version ever shipped satisfies the check.

## Root cause

The literal `1` was written when the constant was also `1`, and the unused-constant pattern (no caller, no test asserting the relationship) let the duplication rot.

1. The fix is trivial in isolation (use the constant) but the module is orphaned and never called; the real decision is whether deployment compatibility belongs in this contract at all (it is a deploy-time tool concern) and whether `REQUIRED_PROTOCOL_VERSION` should match a real minimum (e.g. the protocol version the contract's features actually require).
2. No contract code can realistically be blocked from deployment by a compile-time check, so the function's purpose must be redefined (e.g. a panic-on-incompatible guard in `initialize`/`migrate`) or dropped.
3. The check interacts with `version_negotiation`'s `PROTOCOL_VERSION` — two "protocol version" concepts that must not diverge.

## Acceptance criteria

- [ ] `verify_deployment_compatibility` uses `REQUIRED_PROTOCOL_VERSION` (or the module is removed).
- [ ] The minimum protocol version reflects a real requirement, and a test asserts the relationship.
- [ ] The orphan-file lint (companion issue) covers this module's fate.

## Out of scope

Introducing a real deploy-time gate and the other orphan modules.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/deployment_policy.rs`, `apexchainx_calculator/src/lib.rs` (module list).

ustaxs:
## Description

## Type of Change
- [ ] Documentation update

## Related Issues

## Changes Made
- Closes #442
- Closes #443 
- Closes #444 
- Closes #445 
- 
- 

## Testing

## Checklist

### DIFF
--- a/DeploymentPolicy.md
+++ b/DeploymentPolicy.md
@@ -1,0 +1,31 @@
+Problem
+src/deployment_policy.rs (orphan module — see companion issue) declares the constant but checks a literal:
+
+pub const REQUIRED_PROTOCOL_VERSION: u32 = 1;
+
+pub fn verify_deployment_compatibility(env: &Env) -> bool {
+    let current_protocol = env.ledger().protocol_version();
+    current_protocol >= 1   // <- hardcoded, ignores REQUIRED_PROTOCOL_VERSION
+}
+Consequences:
+
+The constant is decorative: changing REQUIRED_PROTOCOL_VERSION to 2 would have no effect on the check; the guard is a tautology.
+The check is meaningless on every live ledger: Soroban ledgers run protocol versions far above 1 (protocol 20+ is current), so >= 1 is always true; the "deployment compatibility verification" gives operators no protection and no signal.
+The module's promise ("verifies that the target ledger meets minimum version requirements") is unfulfilled: any protocol version ever shipped satisfies the check.
+Root cause
+The literal 1 was written when the constant was also 1, and the unused-constant pattern (no caller, no test asserting the relationship) let the duplication rot.
+
+Why this is architecturally hard
+The fix is trivial in isolation (use the constant) but the module is orphaned and never called; the real decision is whether deployment compatibility belongs in this contract at all (it is a deploy-time tool concern) and whether REQUIRED_PROTOCOL_VERSION should match a real minimum (e.g. the protocol version the contract's features actually require).
+No contract code can realistically be blocked from deployment by a compile-time check, so the function's purpose must be redefined (e.g. a panic-on-incompatible guard in initialize/migrate) or dropped.
+The check interacts with version_negotiation's PROTOCOL_VERSION — two "protocol version" concepts that must not diverge.
+Acceptance criteria
+ verify_deployment_compatibility uses REQUIRED_PROTOCOL_VERSION (or the module is removed).
+ The minimum protocol version reflects a real requirement, and a test asserts the relationship.
+ The orphan-file lint (companion issue) covers this module's fate.
+Out of scope
+Introducing a real deploy-time gate and the other orphan modules.
+
+Getting started
+just test
+Good first files to read: apexchainx_calculator/src/deployment_policy.rs, apexchainx_calculator/src/lib.rs (module list).


usmanimamu17-create (approved):
LGTM 👍

PR #557, opened 2026-08-31
documents insertion order

### LINKED ISSUE #450: get_config_bundle and get_config_snapshot omit custom severities: bootstrap consumers never receive custom configs
## Problem

`get_config_snapshot` (`src/lib.rs`/`src/config.rs`) builds entries from `canonical_severities` only, and `get_config_bundle` composes it with `get_result_schema`:

```rust
pub fn get_config_snapshot(env: Env) -> Result<SLAConfigSnapshot, SLAError> {
    ...
    for severity in Self::canonical_severities(&env) { ... }
```

Custom severities live in a separate endpoint (`get_custom_config_snapshot`), which the bundle does not call. The bundle doc says it is "the recommended way for backend consumers to bootstrap their configuration cache in one RPC" and to "Use the snapshot for SLA evaluation parameters".

Consequences:

- **A backend bootstrapping from the bundle cannot evaluate custom-severity outages**: `calculate_sla` accepts custom severities, but a consumer whose cache came from `get_config_bundle` has no parameters for them; first evaluation either fails with `ConfigNotFound` or, worse, uses a stale default.
- **The one-RPC bootstrap promise excludes a whole feature**: #93 deployments must issue a second RPC (`get_custom_config_snapshot`) and stitch two snapshots with different schemas — the exact fragmentation the bundle exists to prevent.
- **The hash-based cache check compounds the gap**: `compute_config_version_hash` ignores custom configs (companion issue), so a cache refreshed on hash change never updates for custom-severity edits.

## Root cause

The bundle was built from the pre-#93 snapshot; the custom snapshot endpoint was added alongside instead of integrated.

1. `ConfigBundle` is `#[contracttype]` and stability-guarded; adding a `custom_snapshot` field is additive (safe per the append-only rule) but requires updating the guardrail's `canonical_field_counts` (companion issue) and the bundle's consumers.
2. Alternatively, `SLAConfigSnapshot` itself could carry a `custom_entries` vector — a change to a widely-consumed type — so the design decision is where custom entries belong in the type hierarchy.
3. The cache-invalidation semantics (which hash changes) must be fixed in the same release for the bundle to be trustworthy (companion issue on the hash).

## Acceptance criteria

- [ ] A consumer that bootstraps from the bundle can evaluate every severity `calculate_sla` accepts.
- [ ] The bundle's snapshot (or a documented companion) includes custom severities.
- [ ] Cache refresh via `get_config_version_hash`/`get_last_config_update` reflects custom-severity changes.

## Out of scope

The hash exclusion itself (companion issue) and a merged snapshot endpoint.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`get_config_snapshot`, `get_config_bundle`), `apexchainx_calculator/src/config_bundle.rs`.

### LINKED ISSUE #451: load_config returns ConfigNotFound for a missing custom severity while get_custom_severity returns SeverityNotInSet: one condition, two error codes
## Problem

Two public methods report the same condition — "this custom severity was never registered" — with different errors:

```rust
// src/lib.rs, load_config (used by calculate_sla, get_config, calculate_sla_view)
custom.get(severity.clone()).ok_or(SLAError::ConfigNotFound)

// src/lib.rs, get_custom_severity
custom.get(severity).ok_or(SLAError::SeverityNotInSet)
```

`SeverityNotInSet` (code 18) exists specifically for "Custom severity referenced but not registered"; `ConfigNotFound` (code 4) is documented as "No configuration found for the given severity".

Consequences:

- **Backends must special-case by endpoint**: `error_responses.rs` ships `is_config_not_found` and `is_severity_not_in_set` helpers, but the same lookup via `get_config` vs `get_custom_severity` returns different codes, so error-handling logic keyed to one code silently misses the other path.
- **Failure taxonomy is inconsistent**: `docs/FAILURE_TAXONOMY.md` classifies codes; a caller that treats `ConfigNotFound` as "retry later" and `SeverityNotInSet` as "programming error" will misclassify the same underlying state depending on which method it called.
- **The `dup_input`/replay paths inherit the ambiguity**: `calculate_sla` with an unregistered custom severity returns `ConfigNotFound` while the error catalog implies `SeverityNotInSet` is the custom-severity signal.

## Root cause

`load_config`'s fallback to the custom map was written with `ConfigNotFound` (matching the canonical path's error), while the dedicated `get_custom_severity` endpoint used the purpose-built `SeverityNotInSet`; the two were never aligned.

1. Aligning them is an error-code behavior change: `calculate_sla`/`get_config` callers currently see `ConfigNotFound` for missing custom severities; switching to `SeverityNotInSet` changes the failure surface for live consumers and must be coordinated with the failure taxonomy and the error-catalog companions.
2. `load_config` is the shared lookup for both canonical and custom paths, so the fix must preserve `ConfigNotFound` for canonical (where it is correct) while selecting `SeverityNotInSet` for custom — a lookup-shape decision.
3. Tests currently assert `ConfigNotFound` on the custom path (e.g. via `calculate_sla_view` with unknown custom severity); those assertions must be updated in lockstep.

## Acceptance criteria

- [ ] A missing custom severity returns the same error code through every public path.
- [ ] `error_responses.rs` helpers and the failure taxonomy match the implemented codes.
- [ ] Tests cover the missing-custom-severity case through `get_config`, `get_custom_severity`, and `calculate_sla`.

## Out of scope

The error-catalog single-source-of-truth (companion issue) and adding new error variants.

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`load_config`, `get_custom_severity`), `apexchainx_calculator/src/error_responses.rs`.

### LINKED ISSUE #452: set_custom_severity skips validate_cross_severity_penalty_ordering: custom penalties can invert the ordering canonical configs enforce
## Problem

`set_config` enforces severity progression via `validate_cross_severity_penalty_ordering`, but `set_custom_severity` does not:

```rust
// set_config (src/lib.rs)
Self::validate_cross_severity_penalty_ordering(&env, &severity, penalty_per_minute)?;

// set_custom_severity (src/lib.rs) — only general bounds
Self::validate_general_bounds(threshold_minutes, penalty_per_minute, reward_base)?;
```

The ordering rule (#92) guarantees "higher-severity penalties are never lower than lower-severity ones" among canonical tiers; a custom severity — which `calculate_sla` evaluates exactly like a canonical one — can carry any penalty the general bounds allow, including one that inverts the canonical ladder (e.g. a custom "urgent" severity with penalty 10,000 while critical is capped lower).

Consequences:

- **The incentive model can be inverted**: if an operator treats a custom severity as more severe than critical (a common naming choice), its penalty can undercut critical's, making the "higher severity costs less" outcome the #92 check exists to prevent.
- **The ordering guarantee is tier-scoped without saying so**: the doc for #92 describes a progression rule that applies only to canonical severities; nothing documents that custom severities are exempt, so the guarantee reads broader than it is.
- **Duplicate-detection and recalc accounting treat custom severities as first-class** (via `load_config`), so the config surface is uniform for evaluation but not for validation — an inconsistent trust boundary.

## Root cause

`set_custom_severity` (#93) was implemented with general bounds only; the ordering check was not generalized to handle a severity set that grows beyond the canonical four.

1. `validate_cross_severity_penalty_ordering` is built around the canonical ladder (`canonical_severity_index`, next-higher/next-lower in a fixed 4-array); extending it to custom severities requires defining where a custom severity sits in the ordering (by threshold? by declared "tier"? symbol order?) — a semantic decision, not a mechanical one.
2. Custom severities have no fixed position, so the ordering check needs a total order over severity symbols (e.g. penalty must not exceed the nearest canonical tier's) that is stable and documented.
3. The general bounds (1..=10000) were chosen to allow custom flexibility; tightening the check interacts with that decision and may reject previously-accepted configs — a config-change behavior.

## Acceptance criteria

- [ ] Custom severity penalties cannot invert the documented severity progression (or the exemption is explicitly documented).
- [ ] The ordering rule's scope (canonical vs. custom) is documented in one place.
- [ ] Tests cover a custom severity attempting to undercut critical's penalty.

## Out of scope

Threshold-based ordering (only penalties are ordered today) and the duplicate bounds logic (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`set_custom_severity`, `validate_cross_severity_penalty_ordering`).

### LINKED ISSUE #453: Default config triples are duplicated between initialize and init_missing_storage_defaults: a default change requires two coordinated edits
## Problem

The four canonical default configs are written in two places in `src/lib.rs` with identical literals:

```rust
// initialize()
configs.set(symbol_short!("critical"), SLAConfig { threshold_minutes: 15, penalty_per_minute: 100, reward_base: 750 });
configs.set(symbol_short!("high"), SLAConfig { threshold_minutes: 30, penalty_per_minute: 50, reward_base: 750 });
... // medium, low

// init_missing_storage_defaults() — byte-for-byte the same four sets
if !inst.has(&CONFIG_KEY) {
    let mut configs = Map::<Symbol, SLAConfig>::new(env);
    configs.set(symbol_short!("critical"), SLAConfig { threshold_minutes: 15, penalty_per_minute: 100, reward_base: 750 });
    ...
}
```

Consequences:

- **A default change ships inconsistently**: editing the defaults in `initialize` (what fresh deployments get) without updating `init_missing_storage_defaults` (what migrations backfill) yields different configs for new vs. migrated contracts — a silent divergence between deployment paths.
- **The migration path is where inconsistency bites hardest**: `migrate()` v0→v1 calls `init_missing_storage_defaults` for contracts that predate keys; those contracts would get stale defaults while new deployments get the new ones.
- **Reviewers must diff two ~20-line blocks to verify a default change**: the duplication defeats the review process the storage-key maintenance note already warns about.

## Root cause

`init_missing_storage_defaults` was written by copying the `initialize` block; no shared constant exists for the default config triple.

1. The fix is a shared `fn default_configs(env) -> Map<Symbol, SLAConfig>` used by both sites — mechanically simple, but it must be verified that `initialize` and the migration backfill are the only two writers (they are) and that the shared helper's order/values are pinned by tests.
2. Any behavioral change to defaults is a product decision (the values affect every calculation); the refactor must be behavior-neutral and covered by snapshot tests that would catch an accidental value change.
3. The same duplication pattern exists for the stats/telemetry initialisation blocks (`SLAStats { 0, 0, 0, 0 }` appears in both), so the fix should consider consolidating the whole initial-state block.

## Acceptance criteria

- [ ] Default config values are defined once and used by both `initialize` and `init_missing_storage_defaults`.
- [ ] A snapshot test pins the default configs so a value change is intentional and reviewed.
- [ ] Fresh-deploy and migrated-contract configs are identical after initialization.

## Out of scope

Making defaults configurable on-chain and the validate_config/validate_general_bounds duplication (companion issue).

## Getting started

```bash
just test
```

Good first files to read: `apexchainx_calculator/src/lib.rs` (`initialize`, `init_missing_storage_defaults`).

bilkee:
## Description

## Type of Change
- [ ] Documentation update

## Related Issues

## Changes Made
- Closes #450
- Closes #451 
- Closes #452 
- Closes #453 
- 
- 

## Testing

## Checklist

### DIFF
--- a/Docs-Order.MD
+++ b/Docs-Order.MD
@@ -1,0 +1,32 @@
+Problem
+get_config_snapshot (src/lib.rs/src/config.rs) builds entries from canonical_severities only, and get_config_bundle composes it with get_result_schema:
+
+pub fn get_config_snapshot(env: Env) -> Result<SLAConfigSnapshot, SLAError> {
+    ...
+    for severity in Self::canonical_severities(&env) { ... }
+Custom severities live in a separate endpoint (get_custom_config_snapshot), which the bundle does not call. The bundle doc says it is "the recommended way for backend consumers to bootstrap their configuration cache in one RPC" and to "Use the snapshot for SLA evaluation parameters".
+
+Consequences:
+
+A backend bootstrapping from the bundle cannot evaluate custom-severity outages: calculate_sla accepts custom severities, but a consumer whose cache came from get_config_bundle has no parameters for them; first evaluation either fails with ConfigNotFound or, worse, uses a stale default.
+The one-RPC bootstrap promise excludes a whole feature: #93 deployments must issue a second RPC (get_custom_config_snapshot) and stitch two snapshots with different schemas — the exact fragmentation the bundle exists to prevent.
+The hash-based cache check compounds the gap: compute_config_version_hash ignores custom configs (companion issue), so a cache refreshed on hash change never updates for custom-severity edits.
+Root cause
+The bundle was built from the pre-#93 snapshot; the custom snapshot endpoint was added alongside instead of integrated.
+
+Why this is architecturally hard
+ConfigBundle is #[contracttype] and stability-guarded; adding a custom_snapshot field is additive (safe per the append-only rule) but requires updating the guardrail's canonical_field_counts (companion issue) and the bundle's consumers.
+Alternatively, SLAConfigSnapshot itself could carry a custom_entries vector — a change to a widely-consumed type — so the design decision is where custom entries belong in the type hierarchy.
+The cache-invalidation semantics (which hash changes) must be fixed in the same release for the bundle to be trustworthy (companion issue on the hash).
+Acceptance criteria
+ A consumer that bootstraps from the bundle can evaluate every severity calculate_sla accepts.
+ The bundle's snapshot (or a documented companion) includes custom severities.
+ Cache refresh via get_config_version_hash/get_last_config_update reflects custom-severity changes.
+Out of scope
+The hash exclusion itself (companion issue) and a merged snapshot endpoint.
+
+Getting started
+just test
+Good first files to read: apexchainx_calculator/src/lib.rs (get_config_snapshot, get_config_bundle), apexchainx_calculator/src/config_bundle.rs.
+
+


usmanimamu17-create (approved):
LGTM, thanks!

